PR #20, opened 2021-02-02
Fs quench

f-shaker:
Update quenching Box constants and added a couple of unit tests for the quenching module


soleti (approved):
Looks good, thank you for looking into this

PR #29, opened 2021-08-25
Speed up dumpTree.py for larger files

peter-madigan:
Using a list to collect sub-arrays prior to concatenation scales much better for large-ish (~10k) EDepSim files.


peter-madigan:
@soleti do you know why the docs failed? I didn't touch anything that should impact them...

soleti:
@peter-madigan Uhm, weird. Can it have something to do with the new `larpix` requirements? It looks like it's failing to install `bitarray`?

peter-madigan:
Ah could be. I should probably relax most the the requirements, I'll do
that tomorrow.

soleti:
I think I fixed and in any case is not related to this PR

PR #32, opened 2021-09-09
Upgrade/improve cli loop memory

peter-madigan:
Transfers simulation results off of GPU after each batch for better memory performance with larger files. I did not notice any performance degradation with this change.

Also adds a test for the simulate_pixels.py script.


peter-madigan:
Removed cli test since cupy is not available

soleti (approved):

PR #36, opened 2021-10-19
Upgrade/add discriminator threshold noise

peter-madigan:
- Adds gaussian noise to the discriminator checks in the front-end simulation.
 - Makes the ``time_padding`` and ``time_window`` configurable from the yamls

Let me know if I should separate these into different pull requests. Both are pretty minimal changes, so I don't foresee merge conflicts.


peter-madigan:
Closed in favor of #37

PR #39, opened 2021-11-04
Light Sim Addition

tairaeli:
Added a light simulation model relying on a light lookup table. Made some additions to a few of the other modules to account for these additions.


larndsim/consts.py
```diff
@@ -186,3 +195,11 @@ def load_detector_properties(detprop_file, pixel_file):
...
     n_pixels_per_tile = len(np.unique(xs)), len(np.unique(ys))
 
+
+# hacky stuff for lightLUT that probably belongs in the module descriptor files
+# Numpy array containing the number of TPCs on the x,y and z axis respectively
+n_tpc = [2,1,1] 
+# Numpy array showing the x,y and z sizes of the modules of the TPC
+ModuleDimension = [670.,2022.414,670.]
+# Numpy array displaying the number of modules in the x,y and z axis
+n_mod = [1,1,1]
```
soleti, line 205: This `consts.py` module is not the best and needs refactoring, but could we avoid adding more variables and instead store them in the YAML configuration file for the detector? I don't think you need to store the number of TPCs or module on each axis explicitly, they can be derived from existing information. Also, this is for Module0 but we want to be able to extend this to ND-LAr easily.

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+Code adapted from https://github.com/PPKoller/ArgonCubeLUTSim/blob/main/module_0/lutSim.py
+"""
+
+import sys, time, math
+import argparse
+import numpy as np
+from math import log, isnan
+from . import consts    
+
+def GetLutGeometry(lut_path): # Finds the max and minimum values of the x,y and z axis and the number of voxels in each direction
```
soleti, line 11: The description of the function goes below with triple quotes """, see https://www.python.org/dev/peps/pep-0257/. Also the name of the functions always start with a lowercase letter, so `getLutGeometry`. I prefer `this_convention` and not `thisConvention` for the name of the functions, but it's a matter of taste, we should just try to be consistent though. For the actual content of the docstring we adopted this format: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html, where you need to describe the argument and the return value of the function.

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+    lut_min = np.array([f['Min'][0],f['Min'][1],f['Min'][2]])
+    lut_max = np.array([f['Max'][0],f['Max'][1],f['Max'][2]])
+    lut_ndiv = np.array([f['NDivisions'][0],f['NDivisions'][1],f['NDivisions'][2]])
+
+    return np.array([lut_min,lut_max,lut_ndiv])
+
+def GetVoxel(pos,lut_geometry): # Determines which voxel is being called based on the position of the edep. Voxels are indexed 0-2911
+
+    (lut_min,lut_max,lut_ndiv) = lut_geometry
+    vox_xyz = np.floor(pos/(lut_max-lut_min)*lut_ndiv).astype(int)+lut_ndiv/2
```
soleti, line 23: I don't quite understand here, you apply the floor function, then cast to integer and then add a float? What's a typical value for `lut_ndiv`?

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return int(tpc_x)%2
+
+def LarndToLUTCoord(pos,lut_geometry): # Converts the LArND-sim coord. system to that of the LUT
+
+    # access LUT geometry
+    lut_min,lut_max,lut_ndiv = lut_geometry
+    
+    # should add another condition to be greater than whatever the thckness of the cathode is 
+    lut_pos = pos + np.array([lut_min[0],220,0])
```
soleti, line 40: What's 220 here? we should avoid hardcoded numbers.

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return (lut_pos)
+
+########################################################################################################
```
soleti, line 44: please remove lines like this

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+
+def calculate(t_data,lut_path,light_dat): # Simulates the number of photons read by each optical channel depending on ewhere the edep occurs. Also indicates the time it takes for a photon to reach the nearest photomultiplier tube (the "fastest" photon)
```
soleti, line 46: Can you use a more descriptive name?

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+    
+    # Loads in LUT root file
+    np_lut = np.load(lut_path)
+    
+    # Obtains lut geometry
+    lut_geometry = GetLutGeometry(lut_path)  
+    
+    # Data containers
+    time = np.full((t_data['dE'].size,consts.n_op_channel*2),20.)
```
soleti, line 55: Same comment as above, what's 20 here?

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+    # Defining number of produced photons (from quencing.py)
+    n_photons = light_dat['n_photons_edep'][:,0]
+
+    # Loop edep positions
+    for dE in range(len(t_data['dE'])):
+
+        # Global position
+        pos = (np.array((z[dE],y[dE],x[dE])))*consts.cm2mm
+
+        sys.stdout.write('\r    current position: ' + str(pos) + ('(%.1f %%)' % ((dE+1)/float(len(t_data['dE']))*100)))
```
soleti, line 72: why are you multiplying by 100? Add a comment or a printout with the units

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+            
+            # Flips op channels if edep occurs in tpc 2
+            if (tpc==1):
+                op_channel = (op_channel+consts.n_op_channel/2)%consts.n_op_channel
+            
+            # Determines the travel time of the "fastest" photon
+            if (T1_dat[entry] < time[dE,int(op_channel)]):  
+                time[dE,int(op_channel)] = T1_dat[entry]
+            
+            if (op_channel % 12) > 5:
```
soleti, line 107: Same as above, we should avoid hardcoded numbers. what is 5? what is 12?

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+            # Determines the travel time of the "fastest" photon
+            if (T1_dat[entry] < time[dE,int(op_channel)]):  
+                time[dE,int(op_channel)] = T1_dat[entry]
+            
+            if (op_channel % 12) > 5: 
+                n_photons_read *= consts.norm_lcm_acl
+                
+            # new index for storing light data the optical channels in the second TPC
+            if tpc == 1:
+                op_channel += 48
```
soleti, line 112: Same as above, what is 48?

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+        # loop voxel entry-list
+        for entry in range(len(lut_vox)):
+
+            op_channel = op_dat[entry]
+            
+            # Calculates the number of photons reaching each optical channel
+            n_photons_read = vis_dat[entry]*n_photons[dE]
+            
+            # Flips op channels if edep occurs in tpc 2
+            if (tpc==1):
```
soleti, line 100: does this still work in a detector with more than one module?

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+
+    (lut_min,lut_max,lut_ndiv) = lut_geometry
+    vox_xyz = np.floor(pos/(lut_max-lut_min)*lut_ndiv).astype(int)+lut_ndiv/2
+    voxel = vox_xyz[2]*lut_ndiv[0]*lut_ndiv[1]+vox_xyz[1]*lut_ndiv[0]+vox_xyz[0]
+
+    return voxel
+
+
+def GetHalfDetCopy(pos): # Determines which TPC the edep takes place in
+    tpc_x = math.floor(pos[0]/(consts.ModuleDimension[0]/2.))+consts.n_mod[0]/2*consts.n_tpc[0]
```
soleti, line 30: can you explain this formula?

larndsim/quenching.py
```diff
@@ -47,3 +47,5 @@ def quench(tracks, mode):
             raise RuntimeError("Invalid recombination value")
 
         tracks[itrk]["n_electrons"] = recomb * dE * consts.MeVToElectrons
+
+        light_dat[itrk,0]["n_photons_edep"] = (dE/consts.Wph - tracks[itrk]["n_electrons"])*consts.ScintPreScale
```
soleti, line 51: what exactly is the shape of `light_dat`? Why do you need to do `light_dat[itrk,0]`?

larndsim/quenching.py
```diff
@@ -15,7 +15,7 @@
 logger.info("QUENCHING MODULE PARAMETERS")
 
 @cuda.jit
-def quench(tracks, mode):
+def quench(tracks, light_dat, mode):
```
soleti, line 18: Please add to the docstring a description of `light_dat`

larndsim/lightLUT.py
```diff
@@ -0,0 +1,119 @@
...
+            
+            if (op_channel % 12) > 5: 
+                n_photons_read *= consts.norm_lcm_acl
+                
+            # new index for storing light data the optical channels in the second TPC
+            if tpc == 1:
+                op_channel += 48
+            
+            # Assigns the number of photons read for each optical channel to an array
+            tphotons[dE,int(op_channel)] += n_photons_read
```
soleti, line 115: My impression about this loop is that it could be replaced by vectorized operations. Is there something I am not seeing that forces you to use `for`?

soleti (commented):
I left some comments in line, maybe about hardcoded numbers and documentation. Thanks!

larndsim/lightLUT.py
krwood: @soleti can I suggest some information like this is put in the README under a "for developers" section?

soleti: Sure, the convention in GitHub is to create a `CONTRIBUTING.md` file. I can do that when I have time (Mu2e CM this week...) or feel free to start one.

DanielMDouglas: Fixed in https://github.com/DUNE/larnd-sim/pull/39/commits/f06c570af59b5d185878076d571cfc8eb1b2b134.  We are happy to use `this_convention` even if `thisConventionIsBetter` ;)

larndsim/lightLUT.py
DanielMDouglas: This is the tpc offset, since the LUT coordinates are centered on the TPC center.  This has been fixed in https://github.com/DUNE/larnd-sim/pull/39/commits/67d16e81845f2109c01cb214f22d26f76e67266d and `consts.tpc_offsets` has been made global.  A better solution is to include this translation in the definition of the LUT itself, which will be done soon.

larndsim/lightLUT.py
DanielMDouglas: I believe this was useful for debugging.  It is removed as of https://github.com/DUNE/larnd-sim/commit/67d16e81845f2109c01cb214f22d26f76e67266d and now the only stdout is the elapsed time in `simulate_pixels.py` (as with the quenching step).

larndsim/quenching.py
DanielMDouglas: `light_dat` right now holds both the number of created photons and the  incident photons per optical channel per edep (shape is `(n_edeps, n_channels)`).  Ideally, we want the number of created photons (which _should_ have shape `(n_edeps)`) to be a part of the tracks database (in the same way as `n_electrons` is currently handled).  `n_electrons` seems to be added but not populated in a previous step in the simulation chain, so it's not clear how to include `n_photons` in this way.

DanielMDouglas: As it is written, we are just using the first channel to store the `n_photons` data.  It would be better for us to store this in another 1-dimensional dataset for now, and eventually add `n_photons` to the `tracks` set.

larndsim/quenching.py
DanielMDouglas: fixed in https://github.com/DUNE/larnd-sim/pull/39/commits/2b23cdbe8b76f7f729a6865e5329eff8bbf2600c

larndsim/consts.py
DanielMDouglas: light sim now gets the voxelization from tpc_borders, so this is removed

larndsim/lightLUT.py
DanielMDouglas: this function now gets the voxel from the global coordinate system with a single int call (see https://github.com/tairaeli/larnd-sim/blob/larnd-branch/larndsim/lightLUT.py#L33)

larndsim/lightLUT.py
DanielMDouglas: now called `calculate_light_incidence`

larndsim/lightLUT.py
DanielMDouglas: this is a default value for arrival time delay.  This is removed for now.

larndsim/lightLUT.py
DanielMDouglas: this logic is used to apply different normalizations depending on whether the op channel is LCM or Arclight.  This is now handled by a list called `op_channel_efficiency` that gets loaded in `consts.py`.  I'm not sure this is really an efficiency, though, and warrants some more investigation.

larndsim/lightLUT.py
DanielMDouglas: 48 optical channels per TPC.  This is now specified from the `module0.yaml` as it is detector-specific.

larndsim/lightLUT.py
DanielMDouglas: It will now!  TPC selection is done by reusing the `pixel_plane` variable set in `drifting.py`.  Light calculation step is moved to after this step and before pixel simulation.

larndsim/lightLUT.py
DanielMDouglas: This is used to decide which TPC we are in.  This is removed and we are using `pixel_plane` variable instead.

larndsim/quenching.py
soleti: `n_electrons` is added at `dumpTree.py` stage. It's not ideal but it seems the path of least resistance. Otherwise we could think of including an extra step in `simulate_pixels.py` where we add the fields we need (`n_electrons`, `pixel_plane`, etc.).

larndsim/consts.py
```diff
@@ -112,6 +120,7 @@ def load_detector_properties(detprop_file, pixel_file):
     global ys
     global pixel_pitch
     global tpc_borders
+    global tpc_offsets
```
soleti, line 123: Why are you adding this global variable here? It looks like it's not used anywhere.

cli/simulate_pixels.py
```diff
@@ -94,6 +95,13 @@ def run_simulation(input_filename,
             input_has_trajectories = True
         except KeyError:
             input_has_trajectories = False
+    
+    # Made a separate array to store number of photons generated from quenching.py. Should be added to tracks as another dtype
+    phot_from_edep = np.zeros(len(tracks), dtype = [('n_photons_edep','f4')])
```
soleti, line 100: I would rather avoid creating a new array containing just another field. Could you add a line to `dumpTree.py` that creates this column? In order to use also existing h5 files, you could check if `n_photons_edep` is present, and if not then you can create a copy of `tracks` with this extra column. I prefer this solution than having an independent array. Also, can we just call it `n_photons`? Or is it confusing? I don't have a strong preference.

soleti (commented):
I think the PR looks very good. I just have a couple of minor comments that should be quick to address. Thank you for your patience.

larndsim/consts.py
DanielMDouglas: this was added while we were reworking the voxelization function.  That function now references `tpc_borders`, which is already global, so this is reverted in https://github.com/DUNE/larnd-sim/pull/39/commits/2726498f5bd170582e856636bb0d4f138a4ec047.  Good catch!

tairaeli: We got that fixed in [`2726498`](https://github.com/DUNE/larnd-sim/pull/39/commits/2726498f5bd170582e856636bb0d4f138a4ec047)

cli/simulate_pixels.py
tairaeli: Fixed in https://github.com/DUNE/larnd-sim/pull/39/commits/5bee7527634b636be39c7c0f23f67f47bcd0c0ea . However I should mention that we weren't actually able to test this by rerunning `dumpTree.py`,  so you may want to validate this yourself just to be sure things are working as they should.  Also, I do agree with calling it `n_photons`, it makes more sense in the context of the rest of `tracks` and it should be pretty easy to distinguish between `n_photons` and `n_photons_det` as they're each in their own separate dataset .

PR #40, opened 2021-11-18
Add adc busy delay to trigger logic

peter-madigan:
- Adds a trigger delay (without charge loss) after the successful trigger of the front end to better reflect the true trigger logic.
 - Promotes the dead time to a constant configurable (`RESET_CYCLES`), which does not impact simulation results


soleti:
Overseeded by #41

PR #41, opened 2021-11-18
Upgrade/implement unique channel thresholds

peter-madigan:
Here's the implementation of the lookup table and the independent pixel thresholds. The `CudaDict` makes it straightforward to include other pixel-wise lookup tables in the future, if needed. E.g. pixel pedestals, gain, noise. It changes a few things:
 - the internal larnd-sim pixel id is now a single integer and there are methods for converting from this pixel id <-> x,y,plane
 - inter-event timestamping is now completely handled by `export_to_hdf5`, all other larnd-sim internal timestamps are relative to the current event. Because of this, the event loop in `simulate_pixels.py` now has an implicit assumption that each step in the loop contains at most 1 event
 - it adds the `pixel_thresholds` argument to `fee.get_adc_values`, this is an array of length `pixels_signals.shape[0]` declaring the charge threshold on each pixel

The default behavior of `simulate_pixels.py` is unchanged, but jupyter notebooks that call functions directly may break due to the above changes.

I've also included a couple example threshold files under `larndsim/detector_properties/thresholds/*.npz`. They are pretty small, but I'm not sure how you feel about binary files in the git repo. These could be removed or handled differently, just let me know.


soleti:
Thank you very much Peter, I'll take a look. About the binary files: I already include a few in the examples directory, they have  *.h5 or *.npy extension. Are the *.npz just numpy arrays? If yes, you can change the extension and they should be saved automatically on Git LFS. If not, please add the .npz to the list of extensions managed by Git LFS.

peter-madigan:
> Thank you very much Peter, I'll take a look. About the binary files: I already include a few in the examples directory, they have *.h5 or *.npy extension. Are the *.npz just numpy arrays? If yes, you can change the extension and they should be saved automatically on Git LFS. If not, please add the .npz to the list of extensions managed by Git LFS.

They are a special numpy array file that contains multiple arrays, so I will add them.

soleti (approved):

PR #43, opened 2021-11-19
Upgrade/implement unique channel thresholds

peter-madigan:
Fixes a bug in how pixel ids are created, the reason for #42


peter-madigan:
I think this will be broken after the merge because I needed to add some arguments to the max_pixels

PR #51, opened 2021-12-06
Add finite bandwidth to buffer

peter-madigan:
This PR adds a convolution of the induced current with an exponential-like impulse response function to the front-end simulation.
This may have an impact on the performance since it adds a loop over samples since last reset for each calculation of the integrated signal.


soleti:
From a quick test, the simulation is much slower now (factor of ~10). Is there a way to speed up the calculation using parallelization maybe? Or make the convolution optional?

peter-madigan:
I can make it optional - I don't think you can parallelize it since it depends on when the last trigger occurred

soleti:
I thought you could pre-calculate the trigger times and then allocate the memory that's needed. But it's complicated. For now just make it optional and defaults to False (doesn't run). Same file takes 1079 s vs 144 s for the current version.

soleti (approved):

PR #55, opened 2022-01-12
Add new far-field model

peter-madigan:
Adds an updated far-field model (`response_44-far_field-v2.npy`) that uses a different normalization scheme that produces better results when compared to Module 0 data


soleti:
Can your replace the current `respose_44-far_field.npy` with this new one? Is there a reason why you want to keep both?

peter-madigan:
No reason, I've overwritten the old one.

PR #56, opened 2022-02-25
Fix a few bugs in fee.py

peter-madigan:
Fixes a few bugs / edge cases that I found on a closer read of the self-trigger logic:
 - ADC_BUSY_DELAY should be 9 cycles not 8
 - Signal integration should be able to extend past the end of the current array in the event that the ADC is still busy when end of the array is reached
 - There was a typo in the calculation of the buffer response convolution interval
 - When a hit veto occurs, RESET_NOISE should be used instead of UNCORRELATED_NOISE
 - When a hit veto occurs, no signal should integrate for RESET_CYCLES
 - Allow for a digitization to occur if there is enough charge integrated after the end of the current array
 - Change the 2-tick delay in the timestamps to -2 (rather than +2), since it is the external trigger timestamp that gets the delay not the self trigger timestamp

### DIFF
--- a/larndsim/fee.py
+++ b/larndsim/fee.py
@@ -26,7 +26,7 @@ DISCRIMINATION_THRESHOLD = 7e3 * physics.E_CHARGE
 #: ADC hold delay in clock cycles
 ADC_HOLD_DELAY = 15
 #: ADC busy delay in clock cycles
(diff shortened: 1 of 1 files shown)


soleti:
Thank you for taking the time of reviewing the trigger logic, much appreciated.

PR #59, opened 2022-03-30
Saving true vertex info

soleti:
Saving the true vertices from the `edep-sim` file in the HDF5 format.


krwood (approved):
This looks correct, but initially we called the new data set "eventTruth" instead of (the more consistent) "vertices". Of course it doesn't matter, but just in case there is downstream code expecting the former.

We also might want to concatenate the vertex list before writing to h5 to be consistent with the format of the other data sets? I think you noticed this one having a different structure before and this is the reason.

PR #69, opened 2022-04-18
Lookup-table-based light simulation

DanielMDouglas:
This branch contains modifications to the lookup-table-based light simulation that includes time distribution, optical detector response, and output formatting.  Data comparisons & tuning are still needed.


peter-madigan:
Closing in lieu of #70

PR #70, opened 2022-04-19
Lookup-table-based light simulation

peter-madigan:
Same as PR #69 except with a few revisions based on review comments.

Also cleaned up merge conflicts.


cli/dumpTree.py
```diff
@@ -175,8 +210,8 @@ def dump(input_file, output_file):
             trajectories[iTraj]["pxyz_end"] = (end_pt.GetMomentum().X(), end_pt.GetMomentum().Y(), end_pt.GetMomentum().Z())
             trajectories[iTraj]["xyz_start"] = (start_pt.GetPosition().X(), start_pt.GetPosition().Y(), start_pt.GetPosition().Z())
             trajectories[iTraj]["xyz_end"] = (end_pt.GetPosition().X(), end_pt.GetPosition().Y(), end_pt.GetPosition().Z())
-            trajectories[iTraj]["t_start"] = start_pt.GetPosition().T()
-            trajectories[iTraj]["t_end"] = end_pt.GetPosition().T()
+            trajectories[iTraj]["t_start"] = start_pt.GetPosition().T() / 1e3
```
soleti, line 213: Could we clarify the unit here just for reference?

cli/dumpTree.py
```diff
@@ -140,18 +142,51 @@ def dump(input_file, output_file):
...
+        f.create_dataset('segments', (0,), dtype=segments_dtype, maxshape=(None,))
+        f.create_dataset('vertices', (0,), dtype=vertices_dtype, maxshape=(None,))
+
     segments_list = []
     trajectories_list = []
     vertices_list = []
 
-    for jentry in range(entries):
-        print(jentry)
+    for jentry in tqdm(range(entries+1)):
```
soleti, line 154: Why the +1 here?

cli/simulate_pixels.py
```diff
@@ -221,30 +242,42 @@ def run_simulation(input_filename,
...
     end_idx = len(tracks['eventID']) - 1 - rev_idx
 
+    # create a lookup table for event timestamps
+    event_times = fee.gen_event_times(tot_evids.max()+1, 0)
+
     # We divide the sample in portions that can be processed by the GPU
     step = 1
+    
+    # charge simulation
+    event_id_list = []
```
soleti, line 252: Can you clarify the logic here? I had moved these inside the for in order to save the output after each event, does it now happen every `EVENT_BATCH_SIZE` events?

cli/simulate_pixels.py
```diff
@@ -333,7 +366,7 @@ def run_simulation(input_filename,
 
             RangePush("sum_pixels_signals")
             # Here we combine the induced current on the same pixels by different tracks
-            TPB = (8,8,8)
+            TPB = (1,1,64)
```
soleti, line 369: This is because it's faster?

larndsim/consts/light.py
```diff
@@ -34,5 +83,33 @@ def set_light_properties(detprop_file):
...
         OP_CHANNEL_EFFICIENCY = np.array(detprop['op_channel_efficiency'])
+        
+        tpc_to_op_channel = detprop['tpc_to_op_channel']
+        OP_CHANNEL_TO_TPC = np.zeros((N_OP_CHANNEL,), int)
+        for itpc in range(len(tpc_to_op_channel)):
+            for idet in range(N_OP_CHANNEL):
+                if idet in tpc_to_op_channel[itpc]:
+                    OP_CHANNEL_TO_TPC[idet] = itpc
+
+        LIGHT_TICK_SIZE = float(detprop.get('light_tick_size', LIGHT_TICK_SIZE))
```
soleti, line 94: I see these are now loaded from the YAML, if present, otherwise they assume the default values. However, it would be nice for reference to have these values in the module0 YAML file

larndsim/drifting.py
```diff
@@ -29,7 +29,7 @@ def drift(tracks):
 
         track = tracks[itrk]
 
-        pixel_plane = -1
+        pixel_plane = detector.DEFAULT_PLANE_INDEX
```
soleti, line 32: Thank you for this, much better than -1

soleti (commented):
Looks good to me, apart from some minor comments. Am I right to assume there is no MC truth - light simulation association atm?

cli/dumpTree.py
peter-madigan: This is because I made it so it saves a batch of trajectories every now and then within the loop. For the last iteration, it will always save the batch to the file. I could make this less weird if you'd like (just make a dedicated function for writing to the file and then call it after the loop exits).

cli/simulate_pixels.py
peter-madigan: That's right. It helps speed up the Module 0 simulation a bit, since each event isn't actually that big. That can be adjusted to optimize for different detectors (or made into a script parameter). 

As a side note, I did a little bit of profiling and it now spends about equal time in the "simulation" piece as in the "export_to_hdf5" piece. So, there's a potential x2 speed-up that could be achieved by letting the export to hdf5 run in a separate process. Just something to think about - it would take a fair bit of work.

cli/simulate_pixels.py
peter-madigan: I guess - this was not really intended in this PR. I think you maybe mentioned it a while ago, and I tried running with these parameters. I didn't notice a difference, but yeah this was an accident.

larndsim/drifting.py
peter-madigan: Yes thanks Dan!

peter-madigan:
OK @soleti take a look now. I think I've addressed all your comments.

The only truth information that is saved is the intensity data for each track segment, which does require some interpretation. But it's not obvious to me what other information we would want that isn't just the sample-by-sample truth info (which quickly becomes a ton of mostly irrelevant data). I figure we can start with what we have, and if later on we find out there is additional information that would be good to track, deal with it then.

PR #71, opened 2022-04-22
Allow for arbitrary sipm response models

peter-madigan:
Allows the SiPM response model to be specified in an array rather than just the RLC impulse response.

Also updates the module 0 default parameters and noise model, fixes a bug in the LUT indexing, fixes a bug in the poisson statistics generation, and handles a few edge cases better.


soleti:
It looks like the test failed because numpy (at least in the version used in the test) is not able to load the impulse model?

peter-madigan:
Oh I forgot the check the tests, sorry. I'll figure that out.

peter-madigan:
I can't reproduce this error with a clean install of larnd-sim from that commit, so I'm guessing the issue is related to GitHub / git lfs not working together.

soleti (approved):

PR #73, opened 2022-04-27
Use npz for lightLUT

peter-madigan:
Also skips tracking npz files in git lfs


peter-madigan:
Tested a fresh clone of this and didn't have any issues with git lfs

PR #77, opened 2022-05-09
Add lut smearing enable/disable

peter-madigan:
Also updates the module0 simulation parameters


peter-madigan:
Closed in lieu of #78

PR #78, opened 2022-05-11
Upgrade/add multi module support

peter-madigan:
Adds multi-module support for light simulation


peter-madigan:
Includes #77

larndsim/detector_properties/module0.yaml
```diff
@@ -26,35 +27,38 @@ module_to_io_groups:
   1: [1,2,3,4]
```
soleti, line 27: I guess here it should be `1: [1,2]` right?

peter-madigan: Oh yeah, I guess I missed that

soleti: Can you change it in this PR?

peter-madigan: Done

PR #88, opened 2022-07-11
Prevent out of range indexing

AlexWilkinsonnn:
It is possible to have an invalid time index in the atomic add in `sum_pixel_signals`. This happens fairly frequently but there is no runtime error when the index is only slightly out of range. For an event with a track with a very late start tick (event 15 of `/dune/data2/users/awilkins/extrapolation/nd_cafs/edep/0m/00/FHC.1000000.edep.h5`) I get a cuda runtime error. `cuda-gdb` with memcheck switched on pointed to these atomic adds as the cause.


soleti (approved):

soleti:
Thank you for catching this and taking the time of making the PR

PR #94, opened 2022-08-30
fix bug in trigger timestamp when not using light sim

AlexWilkinsonnn:
I was calculating z from the times and associated trigger times and noticed that the trigger times are much larger than the data packet timestamps. I think the code that makes the trigger packets expects a time relative to the event start time but was getting an absolute time for when the light sim is turned off. Let me know if I've misunderstood something.

Alex

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -527,7 +527,7 @@ def run_simulation(input_filename,
                                output_filename,
                                event_times[np.unique(event_id_list_batch)],
                                is_first_event=last_time==0,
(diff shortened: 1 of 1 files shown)


soleti:
@peter-madigan do you think it's fine to merge this? Unfortunately it looks like I can't add you as a reviewer, not sure why, maybe you're not part of the DUNE GitHub organization

peter-madigan:
Yep looks good to me. Do you know who to contact to get added to the dune GitHub org?

soleti:
@peter-madigan I think the person is @brettviren

PR #95, opened 2022-09-15
a way to avoid duplicate cupy installation

drinkingkazu:
Detect whether `cupy` is already installed or not, and only add `cupy` to requirements if not installed. The issue is that `cupy` may be installed in multiple ways and not all methods let pip to detect it. Plus, `cupy` gets unhappy when two versions are installed. So add a manual check by an import attempt. I actually had this modification for long time in the fork, but I don't want to maintain a fork just for this (and it should be useful and harmless)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-import setuptools
-
 VER = "0.3.1"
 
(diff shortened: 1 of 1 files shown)


peter-madigan:
Closed in favor of #99

Issue #90, opened 2022-08-19
Batching light simulation


peter-madigan:
For very long duration events, the light simulation may take excessively long or run into memory allocation issues because it simulates all time ticks from the first track to the last track in an event. This may become an issue for ND-LAr scale simulations. Similar to how the charge simulation is batched for numbers of tracks, the light simulation should also implement some batching.

PR #96, opened 2022-09-26
Upgrade/implement batching

peter-madigan:
This PR updates the run loop within `simulate_pixels.py` to batch the simulation based on the TPC/Module rather than the track segment index. It introduces a `TPCBatcher` iterator that returns a mask into a track segment array. The batch size can be modified via the `EVENT_BATCH_SIZE`, which sets the number of _TPCs_ to simulate at once (so 2 TPCs = 1 Module).

The light simulation has been updated to only simulate the "active channels" (those of TPCs that have non-zero deposited energy within the current batch), allowing for more performance flexibility when simulating a detector with many TPCs.

Also modified are the tuned light simulation parameters for Module 0, and the creation of a new sub-module `larndsim.util` which contains the `CudaDict` helper class and the new batching classes.


peter-madigan:
Addresses #90

PR #102, opened 2022-10-06
Make simulate_pixels.py more robust against huge tracks datasets

peter-madigan:
- Filters tracks dataset against declared active TPCs
 - Restructures simulate_pixels.py to create a minimal memory footprint prior to filtering tracks dataset
 - Add segment_id to tracks dataset in order to preserve MC truth backtracking

For the 2x2 file that @krwood provided me, I no longer encounter memory overruns on a single CORI 16GB Tesla V100 gpu with a `tracks` dataset containing 20M entries.

The tracks dataset is still loaded up front which takes approx ~30s. The dataset is then filtered on the CPU (~20s), which could probably be sped up by copying the xyz data to the GPU. The subsequent simulation then only takes about another 20s.


krwood:
I see the same improved performance.

PR #105, opened 2022-10-20
Fix inconsistency between mc_packets_assn definition of track_ids and edep-sim segment_id

peter-madigan:
PR #102 introduced a bug due to my misunderstanding of the truth information. This PR adds the "segment_id" to the input file / `tracks` array, the intent here is to always be able to uniquely match to the edep-sim segment (even if we filter the input file for only the active regions within larnd-sim). However, in the output `mc_packets_assn` dataset the `track_ids` are meant to represent the index into the `tracks` dataset that appears in the output file (not the unique segment id). This makes it easier and faster to look up the segments associated with a packet. I made the mistake of thinking that the `mc_packets_assn` track id should represent the segment id. I have hopefully made variable names more clear in `simulate_pixels.py`, and fixed the docs.

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -158,7 +158,7 @@ def run_simulation(input_filename,
     with h5py.File(input_filename, 'r') as f:
         tracks = np.array(f['segments'])
         if 'segment_id' in tracks.dtype.names:
-            track_ids = tracks['segment_id']
+            segment_ids = tracks['segment_id']
         else:
             dtype = tracks.dtype.descr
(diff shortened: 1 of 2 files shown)


krwood (commented):
Good catch - and the renaming can help avoid further confusion. To that end, should we also change the name of the `tracks`/`new_tracks`/etc. object to `segment`?

peter-madigan:
I don't think it's necessary since the edep-sim doesn't have a "track" to be confused with. But if it seems like it helps with naming consistency, that could be useful.

krwood (approved):
I agree it's not necessary

PR #106, opened 2022-10-21
Load the light LUT in to device memory once, rather than transfer on each batch

peter-madigan:
The light LUT was accidentally being copied to the GPU on every batch iteration. This wasn't a performance driver, but the latest LUT from Livio is ~5GB, which if you copy on every iteration makes things very slow.

This PR changes the call signature of `sum_light_signals` to include a `op_channel` array which enables keeping the look up table in device memory.

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -195,7 +195,7 @@ def run_simulation(input_filename,
     response = cp.load(response_file)
 
     TPB = 256
-    BPG = ceil(tracks.shape[0] / TPB)
+    BPG = max(ceil(tracks.shape[0] / TPB),1)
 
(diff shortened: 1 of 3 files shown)


krwood:
I tried running a file that I've been working with through simulate_pixels.py and am seeing different output than before these changes. See attached event displays:

example_after_pr
example_before_pr

(These are meant to be a few NuMI spills on the 2x2 geometry.)

krwood:
Yes, it could have an effect in the display as well - it uses the trigger
information to reconstruct 3D position. If triggers are missing, it might
just be reconstructing wonky z-locations.

krwood:
After the above revisions things look as they did before: 

example_after_pr_v2

PR #109, opened 2023-01-22
support eventID separator as a command-line option 

drinkingkazu:
a minor change so that a user doesn't have to locally install larnd-sim nor modify the system path install to use a string value "eventID" as a separator (the default kept as spillID)

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -659,7 +659,7 @@ def run_simulation(input_filename,
     logger.take_snapshot([len(logger.log)])
     print(f"Elapsed time: {end_simulation-start_simulation:.2f} s")
     logger.archive('loop',['loop'])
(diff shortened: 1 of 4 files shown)


drinkingkazu:
Added another piece of change that logs the time and memory usage over simulation steps for understanding the performance in laymans term w/o using nvprof outside.

krwood:
#111 reads in `EVENT_SEPARATOR` and other formerly hard coded configurations through a configuration file instead of from the command line. I would like to merge that in and then revise this PR to get your memory logger in. How does that sound to you?

drinkingkazu:
@krwood config file sounds good. yes please merge the mem logger! sorry i was traveling till April...

drinkingkazu:
Closing to open a new PR for memory logger

PR #112, opened 2023-02-22
Fix bug for overflow values

peter-madigan:
A small bug that I found when investigating the nan values - if a pixel adc value becomes fully saturated, the adc value is clipped at 256. The data type in the output file is a 8bit unsigned int, so these clipped values end up rolling over back to 0. This fix makes it so they are correctly propagated to the output file with an ADC value of 255.


peter-madigan:
Since no one has protested, I'll merge this in

PR #113, opened 2023-02-22
Handle case for 0 charge sum in backtracking

peter-madigan:
@drinkingkazu this should fix the issue with nan values in the packet fractions


peter-madigan:
Also merging this one in

PR #114, opened 2023-03-17
Fix batch_size warning

mjkramer:


krwood:
Whoops sorry Matt. Should have merged this PR into `feature_spillSim` before merging `feature_spillSim` into master. Can you make this PR into master and we'll push it in quickly as a hot fix.

PR #116, opened 2023-03-21
Add tot_evids definition to avoid error

sam-fogarty:
`tot_evids` was removed in a previous commit and causes an error because it is no longer defined. Happens when `EVENT_SEPARATOR='eventID'`. This PR adds `tot_evids` back in.


krwood (approved):
Thanks for catching this, @sam-fogarty

PR #121, opened 2023-03-28
Adding option to specify a gain file

sam-fogarty:
This pull request contains a new feature for specifying a gain file. It is similar to how the pixel thresholds file feature is implemented (but a bit less involved). A new flag in `cli/simulate_pixels.py` is added to allow a npz file to be specified (the same format as pixel thresholds file, has `keys`, `values`, and `default` arrays). `fee.digitize()` is changed to include a gain input parameter. If `pixel_gains_file` is None (default), `fee.digitize()` is called in `cli/simulate_pixels.py` in the `save_results()` function with `fee.GAIN` like normal. If `pixel_gains_file` is specified, then `pixel_gains_lut` is created (like in the case of pixel thresholds). Then a gain ndarray is made to be the same shape as `integral_list` so that we can multiply `integral_list` by a gain ndarray in `fee.digitize()`. Since `integral_list` has a shape of `(unique_pix, fee.MAX_ADC_VALUES)`, each row corresponds to a particular unique pixel id. So each row of `gain_list` will be filled with the the corresponding gain value for that row (pixel id). There is an example notebook in the `examples` directory of looking at 0.1 MeV electrons and setting the channel gains to varying values. This example serves to help validate this feature (but further validation is welcomed). There is also an example of how to create a gain npz file. While this implementation may be sufficient for this kind of feature, any suggestions on how to change/improve it are welcomed.


krwood:
After discussing at a meeting back in April*, the consensus was that this is a nice feature to have available, even though the main users of larnd-sim are not in a position to utilize it quite yet. There was a request for an explicit demonstration that with the appropriate configuration, the simulation remains unchanged. Unfortunately, it also seems that a conflict has developed in the intervening time. If the aforementioned check could be reported here, and the conflict resolved I'd be happy to merge this in before our next test run for 2x2 production.

* https://indico.fnal.gov/event/59205/contributions/263662/attachments/165763/220332/Gain%20file%20feature%20for%20larnd-sim.pdf

sam-fogarty:
@krwood I've made a few modifications, the main ones being the following:

- replaced np.tile with numpy broadcasting in simulate_pixels.py, which should be a better implementation.

- ensured that if a pixel gains file is not specified, the same code is executed as in previous larnd-sim versions (the value of gain will be fee.GAIN by default). Feel free to double-check this before merging.

- had to raise tpb to 512 when loading a pixel threshold or pixel gain file, since I'd get a cuda error invalid value when trying to specify a file for the 2x2. The 2x2 has pixel ids up to 313599 (when all the pixel pitches are the same between modules), and tpb of 256 was too low to allow for this apparently.

I confirmed that `simulate_pixels.py` can be run without errors with or without a pixel gain file specified (testing on 2x2 edep-sim file). If there are other checks someone would like to see before merging, let me know.

PR #135, opened 2023-05-09
Fix incorrect poisson stats in light simulation

peter-madigan:
The stochastic fluctuations in the # of photoelectrons was wildly wrong due to a bug in `calc_stat_fluctuations`. The poisson fluctuations were being simulated directly using the photo-current and not the number of photoelectrons.

### DIFF
--- a/larndsim/light_sim.py
+++ b/larndsim/light_sim.py
@@ -231,8 +231,8 @@ def calc_stat_fluctuations(light_sample_inc, light_sample_inc_disc, rng_states):
     Simulates Poisson fluctuations in the number of PE per time tick.
     
     Args:
-        light_sample_inc(array): shape `(ndet, ntick)`, light incident on each detector
-        light_sample_inc_disc(array): output array, shape `(ndet, ntick)`, number PE in each time interval
+        light_sample_inc(array): shape `(ndet, ntick)`, effective photocurrent on each detector
+        light_sample_inc_disc(array): output array, shape `(ndet, ntick)`, effective photocurrent on each detector (with stocastic fluctuations)
         rng_states(array): shape `(>ndet*ntick,)`, random number states
     """
     idet,itick = cuda.grid(2)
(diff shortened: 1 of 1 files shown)


peter-madigan:
Before
image

After
image

True expectation value for this event = 47.9pe
Overall normalization of the waveforms is unchanged (to within ±sqrt(47.9)=7):
Before: `(light_sample_inc_disc[channel_index] * LIGHT_TICK_SIZE).sum() == 47.119995`
After: `(light_sample_inc_disc[channel_index] * LIGHT_TICK_SIZE).sum() == 57`

mjkramer:
It looks like `light_sim.sum_light_signals` is where we go from PEs to photocurrent? The function's docstring says
```
light_sample_inc(array): output array, shape `(ndet, nticks)`, number of photons incident on each detector at each time tick (propogation delay only)
```
"Number of photons" should be "photocurrent"?

krwood:
Given the 0.01 us digitization in the light readout simulation, Peter's plots above, which are binned in these ticks, should only see a PE/us values at integers/0.01. The before doesn't do this, but the after does, which is good. But the after also takes the expected signal size from essentially right on the expected value to a bit over a ~1sigma fluctuation up. Doesn't seem problematic in and of itself, but it would be encouraging to see a summary of these integrals for a larger sample to check the integral does fall within 1 sigma ~68% of the time after fixes. Even better would be to also make the same plot for the situation before the fix to show it wasn't doing this.

@peter-madigan does this sound like an easy thing for you to spin up?

peter-madigan:
Yeah I could do a quick check just running this even many times with a
different seed to verify a proper poisson, but I won't be able to get to it
for ~2 weeks.

krwood:
@AWh1t3 has done some checks using spill simulation files with and without the changes in this PR (thanks @mjkramer for producing them). @AWh1t3 please feel free to post your findings here.

She may also be able to look into making the above plot, if it isn't too much work to reproduce what @peter-madigan has already done. @peter-madigan maybe you could share your code with @AWh1t3 for her to look into and run if it would be easier on you.

AWh1t3:
Overall, it created an additional trigger for the light simulation, but going event by event, sometimes one or two triggers were removed from a particular event, and sometimes one or two were added.
Screen Shot 2023-05-18 at 12 00 19 PM

For some waveforms, the sum peaks were shifted earlier in the sampling period, but the the height of the peak remained unchanged, and for others the height of the peak changed (very slightly) without altering the peak position

For the particular spill below, light was triggered where it definitely should have been, but hadn’t in the previous code iteration:
Screen Shot 2023-05-18 at 12 04 05 PM

The simulation triggers the light readout using a threshold placed on the sum channels, but I don't recall the value of the current threshold.

The pre-trigger noise has changed slightly, as well, but not a remarkable amount.

AWh1t3:
Like Kevin said, I'd be happy to look into it more!

krwood:
Going to go ahead and merge this into develop, as it certainly seems like an improvement. If any of the additional checks discussed above happen to be run, please do still feel free to comment on the results in this thread.

Issue #137, opened 2023-05-20
Better definition for the "fraction" attribute of "mc_packets_assn"


drinkingkazu:
The current version of the code [a071fd4](https://github.com/DUNE/larnd-sim/commit/a071fd4510988a7bedc524f2bdef78434d37a8c1) computes the fraction by the following code piece in the [fee.py L428-430](https://github.com/DUNE/larnd-sim/blob/main/larndsim/fee.py#L428-L430):
```
if tot_backtracked != 0:
    for itrk in range(current_fractions.shape[2]):
        current_fractions[ip][iadc][itrk] /= tot_backtracked
```
As such, the "fraction" is normalized within the recorded set of segments. The number of segments to be recorded, N, is fixed while the code runs disregard of the number of segments that actually contribute to the subject packet. For instance, the current default is N=5. 

The actual number of segments contributing to the subject packet (say M in this example) can exceed N. Consider a specific case where the packet carries value 10 contributed from M=6 segments with segment index numbers 1 to 6. The configuration sets N=5. Say the contributions in the respective order are 0.01, 0.02, 0.03, 0.04, 0.05, 0.85. In this case, the current code records the following values.

```
"track_ids": [1,2,3,4,5]
"fraction": [0.067, 0.133,  0.200, 0.267, 0.333]
```
Intuitively, the "fraction" is useful to infer how much fraction of packet value (current, charge) each segment contributed. However, because the fraction is normalized within the recorded set of the segments, this becomes impossible when M>=N. Note the case where M=N is included in this situation since we cannot tell whether M=N or M>N from the given set of information. 

Specifically, this leads to three issues:
- Issue A: We cannot detect the fact that there is unaccounted segments contributing to the packet
- Issue B: We cannot quantify the unaccounted contributions, nor the actual contribution values from the recorded segments.
- Issue C: We could miss more relevant segments with larger contribution value (segment 6 in the above example). This is because the N segments are selected by the order of those segments in the input data array.

I propose two changes.
1. Compute and store the fraction with respect to the total contributing charge (i.e. do not normalize only within the recorded segments). This addresses the issue A and B.
2.  Select the N segments to be recorded by the fraction value (calculated according to 1) in the descending order. This addresses the issue C.

Change 1 is to simply modify the noted lines of the code with the following:
```
if q_sum > 0:
    for itrk in range(current_fractions.shape[2]):
        current_fractions[ip][iadc][itrk] /= q_sum
```
(thanks to @soleti )

[...]

PR #138, opened 2023-05-20
Fix the "fraction" definition

drinkingkazu:
This addresses the issue https://github.com/DUNE/larnd-sim/issues/137


soleti:
I think this can safely be merged, but maybe one extra person should independently verify this. @krwood? 
@drinkingkazu, just to double check, you want to add also the 2x2 notebook?

krwood:
@YifanC is going to handle the testing and approving of this PR. IMO one thing that might be nice to see is summed fractional contributions from a sufficiently complex event for (1) before this PR with MAX_TRACKS_PER_PIXEL set high vs. (2) before this PR with MAX_TRACKS_PER_PIXEL set low vs. (3) after this PR with MAX_TRACKS_PER_PIXEL set high vs. (2) after this PR with MAX_TRACKS_PER_PIXEL set low.

I do agree with @drinkingkazu and @soleti that this seems like a nice improvement. We may even want to put in an error or warning message that is triggered when the summed fractional contributions fall short of 100%?

krwood:
Pull Request #140 is very similar to this one, but it also implements a fractional contribution normalization that is in essence `q_sum` but without the noise superimposed on the signal. I suggest we merge #140 and close this one, but please shout if you disagree @drinkingkazu.

PR #140, opened 2023-06-02
Feature t0 vertex

YifanC:


larndsim/fee.py
```diff
@@ -415,19 +443,20 @@ def get_adc_values(pixels_signals,
...
                         current_fractions[ip][iadc][itrk] = 0
                     last_reset = ic
                     continue
 
-                tot_backtracked = 0
-                for itrk in range(current_fractions.shape[2]):
-                    tot_backtracked += current_fractions[ip][iadc][itrk]
+                #tot_backtracked = 0
+                #for itrk in range(current_fractions.shape[2]):
+                #    tot_backtracked += current_fractions[ip][iadc][itrk]
```
krwood, line 455: [nit-picky comment alert] Might as well remove the commented out, legacy code

krwood (approved):
@YifanC and I tested this this new feature by running the simulation on an events with segments ("tracks" in larnd-sim) that saturate the the `MAX_TRACKS_PER_PIXEL` configuration variable. The situation is handled much more sensibly with this PR.

- Before: all of the fractional contributions sum up to 1 by construction
- After: a pixel that has been "saturated" in the simulation would be indicated by the fractional contributions not summing up to 1

This is a very similar solution to Issue #137 as Pull Request #138, but is also robust to the randomness from the noise simulation. Merging this PR should also close both #137 and #138.

PR #141, opened 2023-06-09
Feature lightupdate

liviocali:
Added features:
 - Read LUT dimensions from file
 - Update SiPM gain and thresholds
 - Force light trigger to beam spill (temporary solution)


krwood (changes requested):
It looks like you have some conflicts that need to be resolved here. Also seems like you may have based this feature branch off of an older version - e.g. https://github.com/DUNE/larnd-sim/blob/feature_lightupdate/cli/simulate_pixels.py#L408 is a clear difference wrt what is in develop

krwood (approved):

PR #150, opened 2023-07-25
Light noise scale

AWh1t3:
Screen Shot 2023-07-25 at 9 52 00 AM
Previously, noise output wasn't digitized. These changes apply the same digitization the signals receive, such that min(noise output)/4 =1.


krwood (changes requested):
This is minor, but can you switch the default light noise file simulate_pixels.py will consider if the user doesn't specify this option in the cli? This new default isn't consistent with the other default input files.

AWh1t3:
I've committed the change, do I need to make another pull request?

PR #152, opened 2023-07-26
Fixing some naming/labeling inconsistencies

sam-fogarty:
I noticed some inconsistencies in the naming/labeling of some variables like `eventID`. Some of these inconsistencies have led to an error or two when running a fresh clone of the develop branch. The changes I have made are small and are the following:

1. In `dumpTree.py`, changed some parameter names, like `eventID`, `vertexID`, `parentID`, etc, to match the new definitions described in 2x2_sim (https://github.com/DUNE/2x2_sim/wiki/File-data-definitions). So `eventID`->`event_id`, `vertexID`->`vertex_id`, etc.

2. I saw that despite defining the `event_separator` in `simulation_properties`, `simulate_pixels.py` uses a hardcoded `event_id`. I just changed those instances to `sim.EVENT_SEPARATOR`, like elsewhere in the code

3. In `simulation_properties/singles_sim.yaml`, add missing `tracks_dset_name` and change `event_separator` to `event_id`, like in `simulation_properties/2x2_NuMI_sim.yaml`

4. In `consts/sim.py`, change default `EVENT_SEPARATOR` and `TRACKS_DSET_NAME` to `event_id` and `segments` respectively

`dumpTree.py` and `simulate_pixels.py` run without errors on a miniRun3 spill file with these changes.

### DIFF
--- a/cli/dumpTree.py
+++ b/cli/dumpTree.py
@@ -14,11 +14,11 @@ import glob
 from ROOT import TG4Event, TFile, TMap
 
 # Output array datatypes
-segments_dtype = np.dtype([("eventID","u4"),("vertexID", "u8"), ("segment_id", "u4"),
-                           ("z_end", "f4"),("trackID", "u4"), ("tran_diff", "f4"),
+segments_dtype = np.dtype([("event_id","u4"),("vertex_id", "u8"), ("segment_id", "u4"),
+                           ("z_end", "f4"),("traj_id", "u4"), ("tran_diff", "f4"),
                            ("z_start", "f4"), ("x_end", "f4"),
                            ("y_end", "f4"), ("n_electrons", "u4"),
-                           ("pdgId", "i4"), ("x_start", "f4"),
+                           ("pdg_id", "i4"), ("x_start", "f4"),
                            ("y_start", "f4"), ("t_start", "f4"),
                            ("t0_start", "f8"), ("t0_end", "f8"), ("t0", "f8"),
                            ("dx", "f4"), ("long_diff", "f4"),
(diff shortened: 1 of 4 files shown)


krwood (commented):
Thanks for the PR @sam-fogarty.

The edits to `dumpTree.py` were not included in the PRs for MiniRun4 because we use a different version of this script that lives at https://github.com/DUNE/2x2_sim/blob/develop/run-convert2h5/convert_edepsim_roottoh5.py . It isn't great to have two different versions of this script, and it's worse that they exist in completely separate repositories. Unlike `dumpyTree.py`, we haven't made a new `simulate_pixels.py` for the 2x2 NuMI simulation, but have tried to put in changes we need in a way that is backwards compatible with other simulations. This inconsistency is causing issues, and we really should come up with a better working model. I wonder if @mjkramer has any thoughts on this. At the very least, I think we ought to move the linked converter for 2x2 NuMI simulation back into larnd-sim. Whether it's best to keep up with supporting these two scripts or we unify them with appropriate configuration options isn't clear to me. For now, lets merge in @sam-fogarty's fixes to make dumpTree play nice with simulate_pixels.

The changes to the defaults in `sim.py` and `singles_sim.yaml` look fine.

The changes in simulate_pixels seem unnecessary to me. The use of `event_id` instead of `sim.EVENT_SEPARATOR` in these particular places was not an oversight. Changing the name of the field in the `results_acc` dataset should not have any consequence. It's simply the name of a dataset that is passed around the internal larnd-sim calculations - it doesn't reflect an actual input or output dataset name being read in or written out.

sam-fogarty:
Thanks for the feedback @krwood. I'll revert those changes you referred to in your last comment, I didn't realize these were unnecessary changes. And yes I agree that it's not ideal to that two different versions of the conversion script. Maybe this script could live in the larnd-sim repo, and 2x2_sim could use it since larnd-sim will be installed along with 2x2_sim? But maybe there's good reason to keep a version of it in 2x2_sim, so @mjkramer might have more insight.

sam-fogarty:
@krwood I made the change of reverting sim.EVENT_SEPARATOR to event_id in simulate_pixels.py. I did keep the sim.EVENT_SEPARATOR in one instance of `vertices`, since I think this makes sense given that instances of `tracks` uses sim.EVENT_SEPARATOR (let me know though if you also want this one changed). Though I'm wondering what the use of sim.EVENT_SEPARATOR really is. Is this something we actually need?

krwood:
That is a very fair point. We introduced `sim.EVENT_SEPARATOR` when we were using `event_id` to specify distinct interactions and `spill_id` to specify which interactions are grouped together into a spill. In order to be sure that all activity from a spill was being simulating in the same batch, the spill simulation started to use the `spill_id` to separate the batches of segments. We made it configurable to retain backwards compatibility with, e.g., cosmic simulations. In the 2x2-NuMI simulation, we have since decided to repurpose `event_id` to mean what `spill_id` used to mean, and `vertex_id` to specify distinct interactions (what `event_id` used to mean). And a consequence of that is that all users (AFAIK) are using `event_id` as the `sim.EVENT_SEPARATOR`, making it rather pointless. Given it's already implemented, I'm nonetheless leaning towards keeping the feature in in case we want to utilize it in the future. Although I don't have a very strong preference.

sam-fogarty:
Yeah I agree we might as well keep it, in case we want to use it later.

krwood (approved):

PR #153, opened 2023-08-02
Check YAML tpc offsets against GDML tpc offsets

sam-fogarty:
This PR contains a method to check the TPC offsets (from the detector properties YAML files) against the TPC offsets in the GDML used to make the edep-sim input file. The steps are the following:

1. When `dumpTree.py` is run, a ROOT cpp macro is called (`get_module_offsets.cpp`). This script opens the TGeoManager in the edep-sim ROOT file (if such a TGeoManager exists) and recursively traverses the nodes from the top node in the geometry until it finds the nodes of interest (i.e. of the active LAr). It has to do this to get the global coordinates (the tpc offsets) of the nodes of interest by accounting for translations and rotations along the way. Once the tpc offsets are found, they are returned to `dumpTree.py` where they are saved as an attribute in the `segments` dataset.

2. When running `simulate_pixels.py`, the code retrieves the tpc offsets from the attribute saved in the `segments` dataset and passes them to `detector.py`. There it checks the tpc offsets of the YAML against the GDML offsets. If they do not match, a warning is printed along with the YAML and GDML tpc offsets. Currently, the simulation will still use the YAML offsets even if there is a mismatch, but we can change this if we want. Nevertheless, the warning should be sufficient to alert the user of the mismatch.
 
The code should handle when either 1) no TGeoManager was found in the ROOT file (prints a warning) and 2) it could not find the nodes corresponding to the active LAr. For future geometries, we may (and will) need to add new names for the nodes of interest.
https://github.com/sam-fogarty/larnd-sim/blob/3b83c5948ce5215b23457e12ee264cfe96987f93/cli/get_module_offsets.cpp#L45C1-L53C10

There is one issue, but not with this implementation specifically. Given how the 2x2_sim is setup, the TGeoManager is not retained in the spill file. So we cannot take advantage of this check with 2x2 simulations currently.

Tested on:
1. MiniRun4_1E19_RHC.spill.00000.EDEPSIM_SPILLS.root: No TGeoManager in this file, and the code recognizes this and just prints a warning that no TGeoManager can be found, and that the YAML offsets will not be checked.
2. MiniRun4_1E19_RHC.rock.00000.EDEPSIM.root: Has a TGeoManager, and the code finds the GDML offsets and checks them against the YAML offsets. Currently the YAML and GDML offsets match (the GDML used is [v3](https://github.com/DUNE/2x2_sim/tree/develop/run-edep-sim/geometry/Merged2x2MINERvA_v3) I think). 
3. edep_39Ar_betas_10k.root: A module-0 simulation. Has a TGeoManager, and the YAML and GDML offsets match.

Any suggestions on how to improve this PR are welcomed.

### DIFF
--- a/cli/dumpTree.py
+++ b/cli/dumpTree.py
@@ -12,6 +12,7 @@ from tqdm import tqdm
 import glob
 
 from ROOT import TG4Event, TFile, TMap
+import ROOT
 
 # Output array datatypes
 segments_dtype = np.dtype([("eventID","u4"),("vertexID", "u8"), ("segment_id", "u4"),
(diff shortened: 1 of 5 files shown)


mjkramer:
We've decided to take this very nice piece of work and move it into 2x2_sim.

PR #175, opened 2023-12-15
Feature/limited light backtracking with MAX_MC_TRUTH_IDS = 3 and MC_LIGHT_THRESHOLD=1

marjoleinvannuland:
Very limited light backtracking. Per light detector per tick true photons per segment are saved. The maximum number of segments per light detector per tick is 3. Only light signals larger than 1 pe are accepted. The segments are sorted in order of the number of photons they contribute to a light detector, so we get only the most relevant segments.
NB. these settings cause unpredictable cuda memory errors, setting MAX_MC_TRUTH_IDS=0 will disable the light truth info.

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -713,8 +713,8 @@ def run_simulation(input_filename,
                 sorted_indices = np.zeros((n_light_det, selected_tracks.shape[0]), dtype=np.int32)
 
                 for idet in range(n_light_det):
-                    print(light_inc[:,idet]['n_photons_det'])
-                    sorted_indices[idet] = np.argsort(light_inc[:,idet]['n_photons_det'])[::-1]
+                    sorted_indices[idet] = np.argsort(light_inc[:,idet]['n_photons_det'])[::-1] # get the order in which to loop over tracks
+                    print(sorted_indices[idet])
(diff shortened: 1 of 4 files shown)


YifanC:
Hi @marjoleinvannuland , Thank you for the PR.
Some very minor comments. I hope it's not too late.
1. To be able to utilise the default values in "larndsim/consts/light.py", I would suggest something like following:
```MAX_MC_TRUTH_IDS = detprop.get('max_light_truth_ids',MAX_MC_TRUTH_IDS)```
```MC_TRUTH_THRESHOLD = detprop.get('mc_truth_threshold',MC_TRUTH_THRESHOLD)```
 and maybe also add `MC_TRUTH_THRESHOLD` as a global variable.
2. I would suggest to keep the default values for `MAX_MC_TRUTH_IDS` and `MC_TRUTH_THRESHOLD` as it is in "larndsim/consts/light.py", but add them in the yaml in the simulation_properties. Therefore the parameter adjustment would be in the configuration file rather hardcoded in the script.

I will follow up on the waveform validation plot on slack, but I think with the above two minor comments addressed. It should be ready to go.

krwood (approved):
Looks good by eye. Let's merge in for the MiniRun5_beta production over the weekend, and we can take a look again next week.

PR #180, opened 2023-12-23
Feature auto config

YifanC:


YifanC:
There might be easier ways to include different LArPix geometry in different modules, but restructure larndsim to simulate detector readout module by module seem to be most logical and fundamental, so it can also deal with potential difference in the electric field and the definite differences in 2x2 light readout from different modules, especially when each single module is light tight and charge isolated. However, one difficulty is the light trigger might be carried out across modules. It is mostly addressed in this PR, at least for the beam triggering case, but may still require some work. Attention is needed on this front for future development. Since the heavy-duty part of the simulation (batching) is changed from simulating all modules for an event to one module for an event, perhaps we should investigate the efficiency of the GPU use.

PR #183, opened 2023-12-26
add true event_id to backtracking information

### LINKED ISSUE #177: add true event_id to the backtracking dataset

YifanC:
Adding true event id to the packet mc assn to aid downstream backtracking. The updates are tested and are fairly isolated.


krwood (commented):
The event_id information should be accessible through having the segment(/track)_id, but this does provide easier access. Something to keep in mind in case we want to trim down file size though

PR #192, opened 2024-01-16
Fixed Sync/Timestamp + Packets-->Segments Backtracking

seg188:
- Fixed issuing of timestamp packets (more closely matches real data with {almost} 1:1 matching data:timestamp)
- Fixed issuing of sync packets (incorrect packet ordering and issue frequency) + rollover for sync packet timestamp
- Fixed backtracking for packets-->Segments

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -754,7 +754,7 @@ def run_simulation(input_filename,
         logger.take_snapshot()
 
         track_ids = cp.asarray(np.arange(segment_ids.shape[0], dtype=int))
-
+        segment_ids_arr = cp.asarray(segment_ids)
         # We divide the sample in portions that can be processed by the GPU
         is_first_batch = True
         logger.start()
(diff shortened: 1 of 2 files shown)


krwood (approved):
Looks great! Thanks for catching these subtle bugs Stephen! Would be great to post some of the validation plots here for the record, but this is good to go.

seg188:
Before the changes in this PR, we saw that the validation code was reporting that reconstructed charge events were not ~1:1 with true events:

Screen Shot 2024-01-16 at 10 09 28 AM

This issue was two fold. Firstly, due to a bug in the packet-->segment backtracking where the index of the segments was stored instead of the segment_id from edep-sim, the actual backtracking was giving the wrong event IDs. The second issue was with the sync packet logic--the packets were issue every 1e8 instead of 1e7 due to a mis-conversion of clock ticks to microseconds. The effect of this is most obvious when trying to use the sync to reconstruct the global timestamp (packet timestamp + rollover). The previous behavior is here (note duration is shorted than expected 231 seconds, and timestamps are not monotonically increasing):

Screen Shot 2024-01-16 at 10 20 14 AM

Corrected behavior is here:

Screen Shot 2024-01-16 at 10 15 47 AM

Finally, the corrected behavior associating events:

Screen Shot 2024-01-16 at 10 22 29 AM

PR #195, opened 2024-02-02
zero suppress light backtracking matrix

russellphysics:
Light truth backtracking reshaped to suppress null-valued ticks. N-dimensional arrays saved per spill. Apply masks to quickly filter to light truth of interest. Truth fields: 'event_id', 'det_id', 'track_id', 'pe_current', and 'tick'. This implementation was validated on 2x2 production MiniRun 5 edepsim files, verifying on spill, optical detector, and waveform level that information from original backtracking implementation was accurately preserved with new implementation. Example validation attached.
example2

### DIFF
--- a/larndsim/light_sim.py
+++ b/larndsim/light_sim.py
@@ -622,6 +622,28 @@ def sim_triggers(bpg, tpb, signal, signal_op_channel_idx, signal_true_track_id, 
     
     return digit_signal, digit_signal_true_track_id, digit_signal_true_photons
 
+def zero_suppress_waveform_truth(spill, waveforms_true_track_id, waveforms_true_photons):
+    event_id, det_id, track_id, photons, tick = [[] for i in range(5)]
+    indices = [index for index, x in np.ndenumerate(waveforms_true_track_id) if x!=-1]
+    truth_dtype = np.dtype([('event_id','i4'),('det_id','i4'),('track_id','i8'),('pe_current','f8'),('tick','i4')])
+    for i in range(len(indices)):
+        itrig=indices[i][0]
+        idet_module=indices[i][1]
+        isample=indices[i][2]
+        icontent=indices[i][3]
+        event_id.append(spill)
+        det_id.append(idet_module)
+        track_id.append(waveforms_true_track_id[itrig][idet_module][isample][icontent])
+        photons.append(waveforms_true_photons[itrig][idet_module][isample][icontent])
+        tick.append(isample)
+    truth_data = np.empty(len(indices), dtype=truth_dtype)
+    truth_data['event_id'] = np.array(event_id)
+    truth_data['det_id'] = np.array(det_id)
+    truth_data['track_id'] = np.array(track_id)
+    truth_data['pe_current'] = np.array(photons)
+    truth_data['tick'] = np.array(tick)
+    return truth_data
+
 def export_light_wvfm_to_hdf5(event_id, waveforms, output_filename, waveforms_true_track_id, waveforms_true_photons, i_mod=-1):
     """
     Saves waveforms to output file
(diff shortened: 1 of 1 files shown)


YifanC:
Hi Brooke @russellphysics, thanks for the PR! I renamed some attributes to be consistent with the rest of the output. Hope you are okay with it. In addition, when I tested this, I encountered some issues.

`waveforms_true_track_id.size` will always be greater than 0 as long as [max_light_truth_ids](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/detector_properties/2x2_mod2mod_variation.yaml#L64C1-L64C20) is set to a non-zero value. However, [truth_data](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L667) can still be empty due to the filter function you added. When `truth_data` is empty, the data appending [here](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L686) or [here](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L701) may raise error.

If I modify `waveforms_true_track_id.size` to `truth_data.size` [here](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L684C24-L684C52) and [here](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L699) correspondingly, it would be fine for non module variation case, but again having error in [merge_module_light_wvfm_same_trigger()](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L754C5-L754C41), because the backtracking size for each module is different ([ref](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/larndsim/light_sim.py#L768)).

Another issue is that if module variation is enabled, the `det_id` / `op_channel_id` will always be up to the number of single module light channels. Namely, (0, 95) for 2x2, but this `det_id` / `op_channel_id` is expected to be global channel id (0, 385) in the current setup.

Consider this two, I wonder if it would be easier if the light backtracking filter is run after the waveform merging for module variation cases. I suppose something after [this](https://github.com/DUNE/larnd-sim/blob/1f1860519793363f4d8dc81ce661d26eead4df53/cli/simulate_pixels.py#L1099), which can be used for both module variation and non module variation.

Thanks!

russellphysics:
Thank you, Yifan. We recently encountered the same issue of an empty `truth_data` container in the process of code profiling. We added an if statement for an empty container to evade this issue.

russellphysics:
Regarding the issues on compatibility with running larndsim with module variations enabled, unfortunately we did not validate this code with module variations enabled, rather we only validated with the default configuration in [2x2_sim develop branch](https://github.com/DUNE/2x2_sim/tree/develop). We do not yet have remedy for the first issue and need to take a closer look at the module variation specific function referenced. The final issue noted appears straightforward to remedy by indexing the opdet id based on module under consideration.

larndsim/light_sim.py
```diff
@@ -681,7 +686,7 @@ def export_light_wvfm_to_hdf5(event_id, waveforms, output_filename, waveforms_tr
                     f[f'light_wvfm/light_wvfm_mod{i_mod-1}'].resize(f[f'light_wvfm/light_wvfm_mod{i_mod-1}'].shape[0] + waveforms.shape[0], axis=0)
                     f[f'light_wvfm/light_wvfm_mod{i_mod-1}'][-waveforms.shape[0]:] = waveforms
 
-                    if waveforms_true_track_id.size > 0:
+                    if truth_data.size > 0:
```
marjoleinvannuland: This breaks if truth_data=None

mjkramer:
Superseded by #201

PR #200, opened 2024-02-09
minor correction so mpvmpr (particle generation) can run as the beam …

YifanC:
…config


krwood (approved):
Have not run any checks, but looks good by eye

PR #201, opened 2024-02-15
Light backtracking reshape

YifanC:
Based on feature branch [brussell_light_backtrack](https://github.com/DUNE/larnd-sim/tree/brussell_light_backtrack).

This change stores truth information in such format, `[('trigger_id', 'i4'), ('op_channel_id','i4'), ('tick','i4'), ('event_id','i4'), ('segment_id','i8'), ('pe_current','f8')]`, where `trigger_id`, `op_channel_id` and `tick` correspond to the index of the `light_wvfm` dataset which takes the shape of (#light triggers, #light readout channels, #light samples).

It has been tested on one file, `MiniRun5_1E19_RHC.convert2h5.00666.EDEPSIM.hdf5`, with modular variation activated and deactivated. This test used configuration of setting `MAX_MC_TRUTH_IDS` to 0, 20, 50, 100, 200 and `MC_TRUTH_THRESHOLD` to 1. When `MAX_MC_TRUTH_IDS = 0`, the output file size is about 1GB. With other settings, the output file size is about 1.1GB. 

The run time with respect to the setting `MAX_MC_TRUTH_IDS` can be found in the following graph. Note that the default `larndsim/bin/lightLUT.npz` is used for this test. Using the light LUT's with finer granularity will increase the run time by a few minutes.
run_time_vs_max_track

The plot below shows how many segments contribute to a light waveform per optical channel per trigger. The file size with the different settings indicate that `MAX_MC_TRUTH_IDS = 20` may be sufficient, but on the conservative side, I suggest to set it to 50. The detector property yaml files are adjust accordingly.
max_track_ids_guide

In addition, I'm showing the negative sum of the truth contribution in pe and ADC/4 in trigger 0 channel 16 as an example. The amplitudes do not match. Perhaps I'm comparing the wrong quantity. It needs to be looked into further. Other than that I think this branch is ready to be merged.
MR5_666_tirg0_opchannel16

### DIFF
--- a/larndsim/light_sim.py
+++ b/larndsim/light_sim.py
@@ -622,25 +622,25 @@ def sim_triggers(bpg, tpb, signal, signal_op_channel_idx, signal_true_track_id, 
     
     return digit_signal, digit_signal_true_track_id, digit_signal_true_photons
 
-def zero_suppress_waveform_truth(spill, waveforms_true_track_id, waveforms_true_photons):
-    event_id, det_id, track_id, photons, tick = [[] for i in range(5)]
+def zero_suppress_waveform_truth(evt_id, waveforms_true_track_id, waveforms_true_photons):
+    event_id, op_channel_id, segment_id, pe_current, tick = [[] for i in range(5)]
     indices = [index for index, x in np.ndenumerate(waveforms_true_track_id) if x!=-1]
-    truth_dtype = np.dtype([('event_id','i4'),('det_id','i4'),('track_id','i8'),('pe_current','f8'),('tick','i4')])
+    truth_dtype = np.dtype([('event_id','i4'),('op_channel_id','i4'),('segment_id','i8'),('pe_current','f8'),('tick','i4')])
     for i in range(len(indices)):
         itrig=indices[i][0]
         idet_module=indices[i][1]
         isample=indices[i][2]
         icontent=indices[i][3]
-        event_id.append(spill)
-        det_id.append(idet_module)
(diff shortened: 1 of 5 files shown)


marjoleinvannuland:
I think the missing amplitude can be explained by the MC_TRUTH_THRESHOLD value

larndsim/light_sim.py
```diff
@@ -622,30 +622,66 @@ def sim_triggers(bpg, tpb, signal, signal_op_channel_idx, signal_true_track_id,
     
     return digit_signal, digit_signal_true_track_id, digit_signal_true_photons
 
-def export_light_wvfm_to_hdf5(event_id, waveforms, output_filename, waveforms_true_track_id, waveforms_true_photons, i_mod=-1):
+def zero_suppress_waveform_truth(waveforms_true_track_id, waveforms_true_photons, i_evt, i_trig, i_mod=-1):
+    """
+    Filter empty light waveform backtracking which is filled with the default value '-1', and flatten the backtracking info to 1D array.
+
+    Args:
+        i_evt(int): event id
```
marjoleinvannuland, line 630: Which event_id is this? The spill id?

YifanC: The `event_id` refers to a true event. To be inclusive, it could be a spill, a cosmic event or a particle bomb/gun event. The use of `event` seem to be consistent from larnd-sim and onwards.

YifanC:
@marjoleinvannuland This is the same channel same trigger setting the pe threshold to 0. I think it still doesn't match.
MR5_666_tirg0_opchannel16_0pe

krwood (approved):

PR #203, opened 2024-02-16
removes double swap of truth coordinates before writing to file

krwood:
larnd-sim uses a different internal coordinate system than edep-sim. 

Below shows the previous results for truth information (red) and reconstructed hit location (blue). Beam is supposed to be ~ in the z-direction, which suggested that the truth information was not being swapped properly.

image

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -1106,17 +1106,15 @@ def run_simulation(input_filename,
     if light.LIGHT_SIMULATED and mod2mod_variation:
         light_sim.merge_module_light_wvfm_same_trigger(output_filename)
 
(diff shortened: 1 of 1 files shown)


krwood:
Refer to this figure, which has the correct axes labels...

image

mjkramer (approved):
Yeah, we previously wrote the truth info before the main simulation loop, so we needed three calls to `swap_coordinates`, but now that the truth info is written at the end, we just need two. (I guess if the truth were written at the very beginning of `simulate_pixels`, we'd just need one call!)

marjoleinvannuland (approved):

YifanC:
@mjkramer I think they are currently filled at the end because `vertices` and `light_dat` are filled during the simulation... Not a very strong reason though

PR #204, opened 2024-02-17
move the swap coordinate before the data writing

YifanC:
Before:
Screenshot 2024-02-16 at 6 46 29 PM
After with modular variation:
Screenshot 2024-02-16 at 6 41 36 PM
After without modular variation:
Screenshot 2024-02-16 at 6 40 19 PM
Maybe we should revisit the active volume check...


krwood (approved):
Thanks!

PR #207, opened 2024-02-27
fix trigger forwarding issue for non modular variation setups

YifanC:


larndsim/fee.py
```diff
@@ -67,10 +67,10 @@
...
 #: Average time between events in microseconds
 EVENT_RATE = 100000 # 10Hz
-#: Beam trig io group
-BEAM_TRIG_IO = 1
-#: Threshold/LRS trig io group
-THRES_TRIG_IO = 2
+if light.LIGHT_TRIG_MODE == 0: #: Threshold/LRS trigger
+    TRIG_IO = 2
+elif light.LIGHT_TRIG_MODE == 1: #: Beam trigger
+    TRIG_IO = 1
```
mjkramer: This works till someone years later adds an `import fee` somewhere before `set_light_properties` has run. Maybe `def get_trig_io()` instead?

YifanC: @mjkramer I cannot think two steps ahead...and am not sure if my brain is always online... please check again with the new commit

mjkramer (approved):

PR #209, opened 2024-02-28
fix a clear mismatch...

YifanC:


cli/simulate_pixels.py
```diff
@@ -711,8 +711,8 @@ def save_results(event_times, is_first_batch, results, i_trig, i_mod=-1, light_o
             logger.take_snapshot()
             logger.archive('light')
 
-            # Prepare the light waveform padding (only necessary for module variation)
-            if mod2mod_variation:
+            # Prepare the light waveform padding
+            if light.LIGHT_SIMULATED and (light.LIGHT_TRIG_MODE == 0 or light.LIGHT_TRIG_MODE == 1):
```
krwood, line 715: Aren't LIGHT_TRIG_MODE == 0 or LIGHT_TRIG_MODE == 1 the only supported options? Maybe this can just be `if light.LIGHT_SIMULATED:` ?

krwood (approved):
I think the second compound OR statement after the first AND is necessarily TRUE all the time. Perhaps we just remove?

cli/simulate_pixels.py
YifanC: yes, currently. When the whole light detector system is triggered simultaneously. I was thinking in future if we trigger each module/tpc separately then you don't want this padding...

mjkramer (approved):

PR #211, opened 2024-03-04
`NDLAr` Specific `*.yaml` Changes

alexbooth92:
Bringing these changes into `develop` before branching off again for next NDLAr MiniProd - just affects NDLAr specific yamls, should have no impact on the 2x2 workflow.

The big diff of the `simulation_properites` yaml comes mostly from a reordering.


krwood (approved):
Agreed. Everything looks good to me

PR #218, opened 2024-04-25
Update pixel pitch for Module 2

jaafar-chakrani:


krwood:
This is a band aid that solves the issues we were seeing with the module 2 offsets reported by @russellphysics https://indico.fnal.gov/event/64244/contributions/288617/attachments/176623/240114/minirun5_spatial_offset___2x2_simulation_calibration___8april2024.pdf

@jaafar-chakrani found the discrepancy with how [ndlar_flow](https://github.com/DUNE/ndlar_flow) and larnd-sim calculate pixel position, which is sketched out here https://indico.fnal.gov/event/64542/contributions/290387/attachments/177430/241559/20240429_GeometryBugFix.pdf along with the validation of the agreement when adjusting the pixel pitch on the last page.

Eventually, we should change the larnd-sim pixel position calculation to be like what is done in ndlar_flow, using the center position for the tiles rather than assuming they are all adjacent.

Issue #216, opened 2024-03-15
larndsim truth information not updated


YifanC:
Brooke reported that in `segments` dataset, `pixel_plane`, `n_electrons`, `n_photons` are not updated. This is due to a mistake that when we filled the truth information, we didn't use the dataset passed to the simulation with the updated information. This should be fixed in the next iteration of MR5.

YifanC:
If we only pass a subset of the segments, ndlar_flow will complain here: https://github.com/DUNE/ndlar_flow/blob/18a2e26f67cbfc8e00898ec99eb0a5bd3476044a/src/proto_nd_flow/reco/charge/raw_event_generator.py#L470
The fix should be updated along side with the flow fix.

YifanC:
We should also understand why 2x2 active volume selection removes segments and why module active volume selection further removes segments.

YifanC:
Using the first 50events of MR5 00999 file for a test: 
In total there are 44873 segments, and 44863 of these are in the larnd defined active volume.
with larndsim commit [ebaa36c](https://github.com/DUNE/larnd-sim/commit/ebaa36c66839d6d6b061572b26ac52ca2637498f) and ndlar_flow commit [909fbe6](https://github.com/DUNE/ndlar_flow/commit/909fbe60cdd6394b268f8ada9787c0f47d808edf), the total number of segments is consistent in larndsim and flow. The maximum of the prompt hit backtracked segment_id in flow is 44833 and in larndsim is 44833. The numbers match, and the hotfix is okay.
Attempting to remove the hotfix with larndsim commit [d28f888](https://github.com/DUNE/larnd-sim/commit/d28f88835a738abea51393eab0b38d098ec7be13) and ndlar_flow commit [909fbe6](https://github.com/DUNE/ndlar_flow/commit/909fbe60cdd6394b268f8ada9787c0f47d808edf). The number of segments in both larndsim and flow are 44863. Flow didn't trip at the place where I stated above in raw_event_generator. However, if you look at packet/ prompt hit backtracked `segment_id`, the maximum from larndsim is 44778, while the maximum from flow is 44833. Clearly there's a problem even flow doesn't complain...

YifanC:
https://github.com/DUNE/ndlar_flow/blob/1ff5456cb04d3b2e4ced50dacc358310f9f31191/src/proto_nd_flow/reco/charge/raw_event_generator.py#L456
This shows we use the segment id in packet assn as indices for the reference between packets and segments

PR #225, opened 2024-05-08
Feature seg truth hotfix

YifanC:


krwood (commented):
Do you have any validations you can show? Reading through by eye, it's very difficult to follow that various `*tracks` datasets, as things seem to have gotten rather disorganized. I don't love the fact that we are `drifting` and `quenching` and `swap_coordinates` different `*tracks` datasets multiple times in different places. There are nested statements that could rewrite one to be the other in under certain conditions... it's gotten very difficult to follow the code and I worry this makes us prone to introducing bugs, especially as we rely on bringing in new developers. I can get more specific and turn this into an issue, but for this PR it would be nice to at least see some validations that the complicated logic is being handled properly for our current configs.

mjkramer (approved):
I agree with Kevin's general sentiment, but this hotfix seems pretty safe to me. It just adds missing info (`n_electrons`, `t_start`, etc.) to the `segments` dataset that we write to disk. A "why do we do this" comment in the code would be nice, though.

YifanC:
@mjkramer @krwood As Matt suggested, I added comments on why we need the hotfix and why it is a hotfix.
It is a fairly isolated change. `all_mod_tracks` are passed in the module loop as `tracks` which have the truth information updated, but not `all_mod_tracks`, we are patching `all_mod_tracks` here because the unsolved flow issue. 
The code is tested to a degree that it verifies the output contains the updated segments.

YifanC:
Oh btw I updated the [Git Issue #216](https://github.com/DUNE/larnd-sim/issues/216) @krwood

krwood (approved):
I will begrudgingly approve the PR, but I’m still not comfortable with the way the code is laid out. Specific to this PR, I really don’t think it’s great that we simulate the quenching and drifting on one segment dataset, and then simulate the pixel responses, and then re-simulate the quenching and drifting in another segment dataset for “back tracking”. We should backtrack to the exact input to the pixel response simulation (especially with the outstanding reproducibility issues). I wouldn’t be surprised if we find the backtracking bugs you and Kazu are encountering in the ml reco training work are related to this.

YifanC:
> I will begrudgingly approve the PR, but I’m still not comfortable with the way the code is laid out. Specific to this PR, I really don’t think it’s great that we simulate the quenching and drifting on one segment dataset, and then simulate the pixel responses, and then re-simulate the quenching and drifting in another segment dataset for “back tracking”. We should backtrack to the exact input to the pixel response simulation (especially with the outstanding reproducibility issues). I wouldn’t be surprised if we find the backtracking bugs you and Kazu are encountering in the ml reco training work are related to this.

Relate this comment to Kevin's Git Issue #228

Issue #227, opened 2024-05-09
Readability of larndsim


YifanC:
I think one of the reason larndsim readability is low is that when we define intermediate data, we uses shape of previous intermediate data, for example [here](https://github.com/DUNE/larnd-sim/blob/4ca8036a5bb2beaff3dff977808f4677dd3bc049/cli/simulate_pixels.py#L930-L932) (instead of using len(unique_pix), we use pixels_signals.shape[0])

Suggestions:
1. Add comments
2. Explicitly define the intermediate data type
3. Enforce consistency check

PR #229, opened 2024-05-14
Feature restore association count to store

YifanC:
This PR mainly tackles that the sync packets backtracking info has "wrong" shape, which breaks the functions of variable `ASSOCIATION_COUNT_TO_STORE` and `MAX_TRACKS_PER_PIXEL`. The next thing is that I notice we append a mix of array and list for backtracking information since the `results` are casted into np.array's at the beginning of `save_results`, but this change is not essential.

### DIFF
--- a/larndsim/fee.py
+++ b/larndsim/fee.py
@@ -21,6 +21,7 @@ from .pixels_from_track import id2pixel
 
 from .consts.units import mV, e
 from .consts import units
+from .detsim import MAX_TRACKS_PER_PIXEL
 
 #: Number of back-tracked segments to be recorded
 ASSOCIATION_COUNT_TO_STORE = 10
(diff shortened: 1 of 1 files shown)


krwood (approved):
Validation plots are always appreciated... looks fine by eye though

YifanC:
@krwood Before this fix, the sync packets backtracking information comes with a length of 2* `ASSOCIATION_COUNT_TO_STORE` while we want it to be `MAX_TRACKS_PER_PIXEL`. (In Git Issue #227 I've complained that we should modify how we make these arrays, so it would be clearer what it represents.) It worked only because we happen to set `ASSOCIATION_COUNT_TO_STORE=10` and `MAX_TRACKS_PER_PIXEL=20`. It will not cast a non-regular shaped list to array unless its datatype is object (https://github.com/DUNE/larnd-sim/blob/40395800933ec9284a0604eba710ad4e2e13f725/larndsim/fee.py#L346-L348). larndsim would not run without a fix like this if `ASSOCIATION_COUNT_TO_STORE` and `MAX_TRACKS_PER_PIXEL` are varied.
It is verified after this, `ASSOCIATION_COUNT_TO_STORE` and `MAX_TRACKS_PER_PIXEL` behave as expected. The current suite of validation plots would not reflect that.

larndsim/fee.py
```diff
@@ -322,9 +322,9 @@ def export_to_hdf5(event_id_list,
                     last_time_tick = time_tick
                     packets.append(TimestampPacket(timestamp=np.floor(event_start_time_list[0] * CLOCK_CYCLE * units.mus/units.s)) ) # s
                     packets[-1].chip_key = Key(io_group,0,0)
-                    packets_mc_evt.append([-1])
-                    packets_mc_trk.append([-1] * (ASSOCIATION_COUNT_TO_STORE * 2))
-                    packets_frac.append([0] * (ASSOCIATION_COUNT_TO_STORE*2))
+                    packets_mc_evt.append(np.array([-1]))
+                    packets_mc_trk.append(np.array([-1] * track_ids.shape[1]))
+                    packets_frac.append(np.array([0] * current_fractions.shape[2]))
```
mjkramer: Any reason we can't just use `MAX_TRACKS_PER_PIXEL` here?

YifanC: We can. I was thinking to do it thoroughly once altogether to enhance the readability of the code and to reduce potential inconsistency

YifanC: and to reduce inconsistency

mjkramer: We can start with small steps. I'd prefer an inconsistent mixture of readability and unreadability rather than consistent unreadability.

PR #230, opened 2024-05-15
Feature signal indexing

YifanC:
1. Corrected an indexing error in sum_pixel_signals
2. Increased `MAX_TRACKS_PER_PIXEL` from 20 to 50, and `MAX_ADC_VALUES` from 20 to 30. For a MiniRun5 file, you would still get a lot more segments than 50 contributing to a pixel.

The run time is ~50min (I'm not sure if it's due to `MAX_TRACKS_PER_PIXEL` and `MAX_ADC_VALUES`). Without light simulation it's ~30min.

This PR needs the change in PR #229. (I could merge then PR... if it's easier...) 
Bug identified by Kazu

### DIFF
--- a/larndsim/detsim.py
+++ b/larndsim/detsim.py
@@ -525,7 +525,7 @@ def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, tr
                     break
 
             if counter < 0:
-                overflow_flag[0] = 1
+                overflow_flag[pixel_index] = 1
 
 
 @cuda.jit
--- a/larndsim/detsim.py
+++ b/larndsim/detsim.py
@@ -469,7 +469,8 @@ def sign(x):
     return 1 if x >= 0 else -1
 
 @cuda.jit
-def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, track_pixel_map, pixels_tracks_signals):
+def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, track_pixel_map, pixels_tracks_signals,
+                      overflow_flag):
     """
     This function sums the induced current signals on the same pixel.
     Converting "signals" from per segment to per pixel ("pixel_signals" and "pixels_tracks_signals")
@@ -489,6 +490,8 @@ def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, tr
             the unique pixels array and the array containing the pixels for each track.
         pixels_tracks_signals (:obj:`numpy.ndarray`): 3D array that will contain the waveforms
             for each pixel and each track that induced current on the pixel.
+        overflow_flag (:obj:`cp.array`): Single-element output array to indicate whether
+            MAX_TRACKS_PER_PIXEL is insufficient
     """
     # itrk goes up to the total number of the segments in this batch
     # ipix goes up to the max number of pixel for any segment
@@ -522,7 +525,7 @@ def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, tr
                     break
 
             if counter < 0:
-                print("More segments per pixel than the set MAX_TRACKS_PER_PIXEL value, ", MAX_TRACKS_PER_PIXEL)
+                overflow_flag[0] = 1
 
 
 @cuda.jit
--- a/larndsim/detsim.py
+++ b/larndsim/detsim.py
@@ -522,7 +522,7 @@ def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, tr
                     break
 
             if counter < 0:
-                print("More segments per pixel than the set MAX_TRACKS_PER_PIXEL value.")
+                print("More segments per pixel than the set MAX_TRACKS_PER_PIXEL value, ", MAX_TRACKS_PER_PIXEL)
 
 
 @cuda.jit
--- a/larndsim/detsim.py
+++ b/larndsim/detsim.py
@@ -542,6 +542,9 @@ def get_track_pixel_map(track_pixel_map, unique_pix, pixels):
             track.
     """
     # index of unique_pix array
+    # although this function could get rid of some segments depending on MAX_TRACKS_PER_PIXEL
+    # the index is with respect to the total number segments in the batch
+    # so when it is translated to "segment_id", it is correct
     index = cuda.grid(1)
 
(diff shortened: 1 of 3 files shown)


larndsim/detsim.py
```diff
@@ -497,23 +504,26 @@ def sum_pixel_signals(pixels_signals, signals, track_starts, pixel_index_map, tr
...
         if pixel_index >= 0:
-            counter = 0
+            counter = -99
             for track_idx in range(track_pixel_map[pixel_index].shape[0]):
-                if itrk == -1:
+                if itrk == -1: # would itrk ever be -1?
                     continue
                 if itrk == int(track_pixel_map[pixel_index][track_idx]):
                     counter = track_idx
+                    if counter >= 0 and itick < signals.shape[2]:
```
krwood, line 516: It looks like this is the crucial difference in this PR (apart from changing `MAX_TRACKS_PER_PIXEL` and `MAX_ADC_VALUES`). Am I right in thinking that not having the `counter >= 0` in this if statement was causing the backtracking errors seen before? And did including this catch expose that we needed to increase the MAX values?

krwood (commented):
Do you have any before and after plots showing the back tracking behavior for one of the edge cases that previously returned questionable associations?

krwood:
>  For a MiniRun5 file, you would still get a lot more segments than 50 contributing to a pixel

Do you have a contributions/pixel distribution we can take a look at to understand where we should set this? I was under the impression we already tuned this to a reasonable value. I sit the NuMI beam simulation where the >50 is coming from, or from particle bombs for mlreco training?

YifanC:
This set of slides should serve as a visual aid for this problem and a validation: https://indico.fnal.gov/event/64745/contributions/291455/attachments/177843/242338/2024_05_14_larndsim_debug_MAX_TRACKS_PER_PIXEL.pdf

YifanC:
Unfortunately, I have seen many many many pixels have more than 50 segments contributing to the signal using MR5 0666 edep input. (The default is set to 20 for MR3-MR5 beta1, as far as I know.)
Tuning `MAX_TRACKS_PER_PIXEL` and `MAX_ADC_VALUES` is not good long term solution. Kazu proposed to rank the contribution in the waveform level before we decide which to keep, and it seems like he is willing to implement it. This could be a solution.

larndsim/detsim.py
YifanC: This set of slides should serve as a visual aid for this problem and a validation: https://indico.fnal.gov/event/64745/contributions/291455/attachments/177843/242338/2024_05_14_larndsim_debug_MAX_TRACKS_PER_PIXEL.pdf

YifanC: The problem is a combination of the default `counter` value and where to fill `pixels_signals` and `pixels_tracks_signals`. Regardless, it's dangerous to set the default value to something physical. 0 and -1 (I think) both can be valid values.

mjkramer: Isn't `track_pixel_map[pixel_index]` supposed to be guaranteed to contain `itrk`, since `pixel_index` comes from `pixel_index_map[itrk]`? As long as this holds, shouldn't `counter` always get set properly, so that the default value doesn't matter?

YifanC: @mjkramer `track_pixel_map[pixel_index]` does not guarantee to contain `itrk`. See the added comments. track_pixel_map.shape[1] goes up to `MAX_TRACKS_PER_PIXEL`, while signals include contribution from all possible segments. For example, if you set `MAX_TRACKS_PER_PIXEL` to 20 and you actually have 25 segments contributing to this pixel. There are some `itrk` value wouldn't find anything in this loop and `counter` remains 0, so the last 5 contributions (not ranked by size) are added to the 0th.
May the pictures here would help
https://indico.fnal.gov/event/64745/contributions/291455/attachments/177843/242338/2024_05_14_larndsim_debug_MAX_TRACKS_PER_PIXEL.pdf

mjkramer: Got it, thanks.

YifanC:
Shall we set `MAX_TRACKS_PER_PIXEL` and `MAX_ADC_VALUES` to 20 and 20, as before?

mjkramer:
I'm seeing normal (~20-minute) run times with `MAX_TRACKS_PER_PIXEL = 50` and `MAX_ADC_VALUES = 30`, although I've disabled the warning printout, as it was really spammy and may have contributed to the slowdown you reported. I also can't see how any of this is related to the light simulation.

YifanC:
Yea.. I guess the print out was not a good idea = = In fact, with the print out, with `MAX_TRACKS_PER_PIXEL = 20` and `MAX_ADC_VALUES = 20` it would jam the run. Shall we do a silent kill (remove the print out completely) and maybe even set it to `MAX_TRACKS_PER_PIXEL = 80` and `MAX_ADC_VALUES = 30` if the memory withhold

mjkramer:
It's still good to have a warning. I've moved the printing outside the kernel. Will see what happens with `MAX_TRACKS_PER_PIXEL = 80`.

mjkramer:
Is there any evidence that `MAX_TRACKS_PER_PIXEL` of 80 preserves significantly more information than 50? In both cases I just see one spill (per module) producing a warning. However, going from 50 to 80 increases the run time from 23 to 28 minutes. I didn't see a substantial difference in peak GPU memory usage (~77GB by eye in both cases... scary).

YifanC:
With MR5 0666, I have seen multiple incidences that it's above 50 contributors despite my initial warning wasn't ideal. I think 77GB memory usage is on the edge... and has huge limitation on which GPU you can run on. (Although when I tested I was using a 40GB memory A100... but I didn't actively monitor the memory use) How much is the memory usage for `MAX_TRACKS_PER_PIXEL = 20` and `MAX_ADC_VALUES = 20`? Probably bump it from `MAX_TRACKS_PER_PIXEL` from 50 to 80 is unnecessary. The question is whether we keep it at a lower number.

YifanC:
@mjkramer Are you okay to merge this in `develop`?

mjkramer:
I see no difference in run time (24 minutes) or GPU memory (51 GB) between (20, 20) and (50, 30) for `MAX_TRACKS_PER_PIXEL` and `MAX_ADC_VALUES`.

PR #236, opened 2024-06-04
Feature seg hotfix removal

YifanC:
Remove the hotfix for the segment updates

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -625,21 +625,20 @@ def run_simulation(input_filename,
     else:
         mod_ids = consts.detector.get_n_modules(detector_properties)
 
-    # Select segments that are in the active volume
     # If mod2mod variation, we load detector properties to get detector.TPC_BORDERS
     # For this purpose, it doesn't matter which pixel_layout to use
     if mod2mod_variation:
         consts.detector.set_detector_properties(detector_properties, pixel_layout[0])
         from larndsim.consts import detector
+
     # Sub-select only segments in active volumes
-    if mod2mod_variation or sim.IF_ACTIVE_VOLUME_CHECK:
-        print("Skipping non-active volumes..." , end="")
-        start_mask = time()
-        active_tracks_mask = active_volume.select_active_volume(all_mod_tracks, detector.TPC_BORDERS)
-        tracks = all_mod_tracks = all_mod_tracks[active_tracks_mask]
-        segment_ids = all_mod_segment_ids = all_mod_segment_ids[active_tracks_mask]
-        end_mask = time()
-        print(f" {end_mask-start_mask:.2f} s")
+    print("Skipping non-active volumes..." , end="")
+    start_mask = time()
+    active_tracks_mask = active_volume.select_active_volume(all_mod_tracks, detector.TPC_BORDERS)
+    tracks = all_mod_tracks = all_mod_tracks[active_tracks_mask]
+    segment_ids = all_mod_segment_ids = all_mod_segment_ids[active_tracks_mask]
+    end_mask = time()
+    print(f" {end_mask-start_mask:.2f} s")
 
     RangePop()                  # prep_simulation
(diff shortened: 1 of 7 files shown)


cuddandr:
Lines that define TPB/BPG and perform the quench and drift are deleted in this commit. Are they somewhere else in simulate pixels now? The commit diff stops showing the rest of the file and it's easier to just ask.

Relevant lines:
``` 
TPB = 256
BPG = max(ceil(all_mod_tracks.shape[0] / TPB),1)
quenching.quench[BPG,TPB](all_mod_tracks, physics.BIRKS)
drifting.drift[BPG,TPB](all_mod_tracks)
```

YifanC:
@cuddandr Yes, they are here in the modular loop. The hotfix is to do it again on the entire set. https://github.com/DUNE/larnd-sim/blob/ebaa36c66839d6d6b061572b26ac52ca2637498f/cli/simulate_pixels.py#L673-L697

cli/simulate_pixels.py
```diff
@@ -617,27 +617,34 @@ def save_results(event_times, is_first_batch, results, i_trig, i_mod=-1, light_o
...
         mod_ids = consts.detector.get_n_modules(detector_properties)
 
+    # Select segments that are in the active volume
+    # If mod2mod variation, we load detector properties to get detector.TPC_BORDERS
+    # For this purpose, it doesn't matter which pixel_layout to use
+    if mod2mod_variation:
+        consts.detector.set_detector_properties(detector_properties, pixel_layout[0])
+        from larndsim.consts import detector
+    # Sub-select only segments in active volumes
+    if mod2mod_variation or sim.IF_ACTIVE_VOLUME_CHECK:
```
mjkramer: Is it necessary for `mod2mod_variation` to secretly override the value of `sim.IF_ACTIVE_VOLUME_CHECK`? This could cause confusion.

mjkramer:
I ran this, and ran the output through flow (`feature_backtrack` branch), and made the validation plots for both. The plots look good.

However, I had to modify both of the plotters because they assumed that `segment_id` could be used as an index into `segments`. That's no longer the case, presumably because of the Active Volume Check (which is now always on when `mod2mod_variation` is enabled; see comment above). The AVC only filtered out a couple of segments, so is there any reason we can't just leave it disabled for now, given that other downstream code might also treat `segment_id` as an index?

YifanC:
@mjkramer I feel strongly that we should avoid using segment_id as index if we can, and should advocate the reconstruction and analysis to not to do that. Another issue is that we should update the boundary in larndsim configuration based on gdml. Directly answering the question why we can't have AVC disabled for `mod2mod_variation`: in the module loop, the input is segments in a particular module selected with AVC function. The sum of these is as AVC activated for the whole detector (double checked). This led to the hotfix.

mjkramer (approved):
OK. We should loudly advise folks that `segment_id` is not an index.

YifanC:
@mjkramer Agreed on loudly informing the changes. I will write a message in the relevant slack channels and mention this in meetings.

cli/simulate_pixels.py
YifanC: Do you mean for this particular if statement or in general? I agree that it's not ideal that `mod2mod_variation` flag overwrites `sim.IF_ACTIVE_VOLUME_CHECK` secretly (it is equivalent to that it is always true). I don't have very good suggestions here. Potentially we can do:
- Update the larndsim detector boundary according to the gdml
- Remove `sim.IF_ACTIVE_VOLUME_CHECK` as an option, and always enable it.

mjkramer: I think doing both would be ideal, but for now, in the interest of getting beta2 out the door, I suggest that we remove the option and proceed. I'll make the change, test, push, and merge, if there are no objections.

YifanC: sounds good! Thank you

PR #253, opened 2024-09-15
Feature config

YifanC:
1. Updated the default light LUT
2. Updated the configuration (moved simulation related parameters to consts.sim)
3. Updated the configuration (moved the parameters in fee.py to consts.detector)
4. Updated Vref (1300 mV -> 1568 mV), Vcm_mV (288 mV -> 478.1 mV), Discriminator_threshold (7ke->5ke) for LArPix (Built based on @krwood's PR #241 )
5. Updated gain (4mV/ke ->4.522mV/ke) | @seg188 promised to provide a reference
6. Match with ndlar_flow [feature_config](https://github.com/DUNE/ndlar_flow/tree/bbff22042b46d27ea02786c2069451e75650838b)

The following features a little comparison between different settings and with respect to the data. The test/validation is done with the first 100 events in `MiniRun5_1E19_RHC.convert2h5.0000999.EDEPSIM.hdf5`. The charge of the hits are calculated based on [this commit](https://github.com/DUNE/ndlar_flow/tree/bbff22042b46d27ea02786c2069451e75650838b) with values indicated below.                     

The two plots compare that the charge distribution with settings as in PR #241 using the yaml configuration or directly change in fee.py. With the caveat that not able to set the random seeds correctly. I think this demonstrate the modification of the configuration setup works.
Hits_Q_5ke_yaml
Hits_Q_7ke_yaml

Tag @seg188 @jaafar-chakrani @krwood @russellphysics to check the logic and comment, especially the part against data.

These plots show the differences with respect to the new and old LArPix setting (running with PR #241 , direct changes in fee.py) The values used in ndlar_flow change correspondingly with the values in larndsim.
Copy the relevant raw code from larnd-sim here:
`adcs = xp.minimum(xp.around(xp.maximum((integral_list * gain + detector.V_PEDESTAL * mV - detector.V_CM * mV), 0)
                                * detector.ADC_COUNTS / (detector.V_REF * mV - detector.V_CM * mV)), detector.ADC_COUNTS-1)`
Copy the relevant raw code from ndlar_flow here:
`(dw / adc_counts * (vref - vcm) + vcm - ped) / gain` 
The threshold change (7ke -> 5ke) increased number of hits by ~25%. The threshold could also lead to lower integrals after triggers, which I think is the sole reason of the change in the lower end of Q. I think the changes in (vped - vcm) and (vref - vcm) cancel out in the simulation and calibration.
The saturation of the charge (upper limit) increases with the new setting. This makes sense as both (vref - vcm) and (vcm - vped) increase.
The new setting also leads to fewer saturated packets setting as (vped - vcm) decreases and (vref - vcm) increases, so in general the adc will be lower.
Hits_Q_old_new_settings

Some screenshots of an example event. The blue is the old setting, and the orange is the new setting. You see more hits in the new setting (lower threshold) and the additional hits appear to be earlier (closer to the anodes) which also makes sense intuitively.
Screenshot 2024-09-09 at 1 44 45 PM
Screenshot 2024-09-09 at 1 46 41 PM

The last part is to show the impact of the gain using the new settings. One small notation that orange uses vcm = 478 mV, while the rest uses vcm = 478.1 mV. The data is the 100-200 events in `packet-0050017-2024_07_09_00_34_36_CDT.FLOW.hdf5` beam-only v4 data.
Some observations:
- Gain in larndsim affects how many packets are saturated; Gain in ndlar_flow affects the charge of the saturated signal. (Makes sense)
- Given the above, comparing orange and red here, the effect of gain, even applied consistently in simulation and calibration, may not be cancelled out. The absolute value of the gain changes the calculated charge.
- Look at the orange and the green. Intuitively it makes sense the green has more saturated hits, and the charge distribution is biased towards higher end.
- Look at the red and the green. They have the exact same larndsim output, and the only difference is the applied gain in ndlar_flow. It makes sense that the green biases towards the higher charge.
- In general the data Q seems to be lower than all the simulation. (for example, green would be biased towards the higher end, if it represents the blue data.)
- I scanned the ~450 events in this data file, not many packets are saturated.
- We know there are negative charges due to ADC time/#hits dependent drift (?) I don't know this explains the last two points.
Hits_Q_5ke_comparison (1)


cli/simulate_pixels.py
```diff
@@ -407,6 +417,18 @@ def save_results(event_times, results, i_trig, i_mod=-1, light_only=False):
         RangePush("load_induction_response")
         response = cp.load(response_file)
         RangePop()
+
+        RangePush("load_pixel_thresholds")
+        if pixel_thresholds_file is not None:
```
krwood, line 422: Does this break if either `pixel_thresholds_file` or `pixel_gains_file`are not set? There doesn't seem to be a catch that sets `pixel_{thresholds,gains}_lut` to their default values if the files are not indicated.

cli/simulate_pixels.py
```diff
@@ -670,6 +678,16 @@ def save_results(event_times, results, i_trig, i_mod=-1, light_only=False):
             response = cp.load(response_file[i_mod-1])
             RangePop()
 
+            RangePush("load_pixel_thresholds")
```
krwood, line 681: Same questions regarding a non-specified pixel_threshold_file. I think it would be useful to support defaults.

cli/simulate_pixels.py
```diff
@@ -1026,23 +1045,26 @@ def save_results(event_times, results, i_trig, i_mod=-1, light_only=False):
...
                 rng_states = maybe_create_rng_states(int(TPB * BPG), seed=rand_seed+ievd+itrk, rng_states=rng_states)
-                pixel_thresholds_lut.tpb = TPB
-                pixel_thresholds_lut.bpg = BPG
-                pixel_thresholds = pixel_thresholds_lut[unique_pix.ravel()].reshape(unique_pix.shape)
+                if pixel_thresholds_file is not None:
+                    pixel_thresholds_lut.tpb = TPB
+                    pixel_thresholds_lut.bpg = BPG
+                    pixel_thresholds = pixel_thresholds_lut[unique_pix.ravel()].reshape(unique_pix.shape)
+                else:
+                    pixel_thresholds = cp.full(pixels_signals.shape[0], detector.DISCRIMINATION_THRESHOLD * consts.units.e)
```
krwood, line 1067: I think this addresses my aforementioned concerns.

larndsim/consts/detector.py
```diff
@@ -74,8 +87,52 @@
 MODULE_TO_TPCS = {}
 TPC_TO_MODULE = {}
 
-ELECTRON_MOBILITY_PARAMS = 551.6, 7158.3, 4440.43, 4.29, 43.63, 0.2053
-
+###################
+# LArPix FEE
+###################
+#: Discrimination threshold in e-
+DISCRIMINATION_THRESHOLD = 7e3 # e-
```
krwood, line 94: I suggest the changing default to 5e3

krwood (approved):
Thanks for the sanity checks. I agree with your assessments - the effect on the dynamic range and saturation levels are as expected. There is also a new light lut that is snuck into this PR. I assume that is an improvement - none of the nice validations you included in the PR address the light simulation. If this was validated elsewhere, good. But if it was an accidental commit, maybe consider removing from this merge and opening a separate PR.

krwood:
Regarding your comparisons to data, I should also note that there is variation in v_ref and v_cm from module-to-module and tile-to-tile that will effect the ADC --> charge conversion. This is mainly due to variations in VDDA (voltage supplied to analog block) from tile-to-tile. Once we extract channel-by-channel gain calibrations from the data, this should be folded in.

krwood:
Also your note of low ADC saturation levels is consistent with the intentions from the operations team and from previous investigations, showing packets with ADC words of >250 at the <0.1% level.

cli/simulate_pixels.py
YifanC: I've only tested the case without setting pixel_{thresholds,gains}_file, where we use the const in the detector yaml, so the answer to your question is yes

larndsim/consts/detector.py
YifanC: This will be overwrite by the yaml. I left it there as we've been running it with 7e3 or so. I think we should have a separate discussion about the threshold, but I don't have strong opinion about the default values.

YifanC:
@krwood 
Re: the light lut, this is just minor correction to match the default LUT (not used in production) with this PR #247. The mainline LUT's are changed in that PR which is already used MR6.
Re: tile-to-tile variation of vref, vcm, I think we can add that support in larndsim, but do we want to run the operational dependent values in big production? also can the variation be fully absorbed by channel-by-channel gain? (I don't see how, maybe I'm just slow= =. Do you mind explaining it?)

YifanC:
Also, I'm concerned that we don't see as clear ADC saturation in data compared to the simulation. I don't know if the instability in ADC through vref, vcm can explain this.

Issue #250, opened 2024-09-11
Check the channel id usage for pixel gain and pixel threshold


YifanC:
Check the channel id usage for pixel gain and pixel threshold (with input files). Make sure we are not only reading the values for the first module.

jaafar-chakrani:
I confirmed during the validation of the pixel thresholds that they are properly set on the expected channels for all modules. The only subtlety that I would like to mention is related to pixel IDs for Module 2:
* For Module 0, the pixel IDs go from `0` to `2 * N_PIXELS[0] * N_PIXELS[1] - 1` with `N_PIXELS = (140, 280)`, i.e. from 0 to 78,399.
* For Module 1, the pixel IDs go from `2 * N_PIXELS[0] * N_PIXELS[1]` to `4 * N_PIXELS[0] * N_PIXELS[1] - 1` with `N_PIXELS` unchanged, i.e. from 78,400 to 156,799.
* For Module 2, the pixel IDs go from `4 * N_PIXELS[0] * N_PIXELS[1]` to `6 * N_PIXELS[0] * N_PIXELS[1] - 1`, except that `N_PIXELS` gets reloaded and now would correspond to `N_PIXELS = (160, 320)`, so they now range from 204,800 to 409,599.
* For Module 3, the pixel IDs go from `6 * N_PIXELS[0] * N_PIXELS[1]` to `8 * N_PIXELS[0] * N_PIXELS[1] - 1`, but `N_PIXELS` gets reloaded back to `N_PIXELS = (140, 280)`, and the range is from 235,800 to 313,599.

So it may look like there is an overlap in pixel IDs between Module 2 and Module 3, but I think it's fine since larnd-sim processes one module at a time independently and this doesn't seem to have any impact. It's only important to keep in mind when we want to make pixel gains/thresholds inputs.

krwood:
IMO, we should try to address this... I understand @jaafar-chakrani's point about having some protection due to module-by-module batching, but that's a configurable parameter in the simulation and we should try to avoid putting in these types of "unspoken assumptions". I.e. in principal, the simulation should work under different batching schemes. If a user changes the batching scheme on their end, they will be susceptible to bugs. Probably worth sorting this out now to avoid any potential confusion down the road.

We really should try and standardize the unique pixel ID scheme and use it here. We have functions to convert from unique pixel IDs to (io_group, io_channel, chip_id, channel_id) and vice versa in many different places, which is potential for bugs (and we've been bitten by them in the past).

PR #263, opened 2025-02-21
Channel-by-channel thresholds extracted from 2x2 data

jaafar-chakrani:
## Implementation
No change to the source code, simply new inputs to pass in `PIXEL_THRESHOLDS_FILE` of `config.yaml`.
2x2 files used to extract thresholds:
- Pedestal data: https://portal.nersc.gov/project/dune/data/2x2/nearline/packet/commission/June2024/noise_assessments_06_05/packet-cold-pedestal-2024_06_05_08_33_21_CDT.h5
- Self-triggered leakage data: https://portal.nersc.gov/project/dune/data/2x2/nearline/packet/commission/June2024/self_trigger_06_10/packet-leakage-threshold-2024_06_10_21_29_09_CDT.h5

Two notebooks are added to this PR:
- `examples/MapPixelIDToUniqueID_2x2.ipynb`: Creates a mapping between a channel unique ID  given by `channel_id + 100*chip_id + 1000*(tile_id + 100*io_group)` and larnd-sim's `pixel_id`
- `examples/Thresholds.ipynb`: Extracts thresholds and save them using this mapping in the expected larnd-sim format

## Validation
1. Impact on the charge distributions
image

2. Mapping from channel ID to larnd-sim's internal pixel ID

To check that the thresholds are properly assigned to the right channels, a larnd-sim file is generated by changing `adc` to `pixel_thresholds[ip]` in: https://github.com/DUNE/larnd-sim/blob/d920a8bbf3aa4b5850884637128ec415ab04afa5/larndsim/fee.py#L670
This allows us to save the thresholds (up to ADC quantization) for each channel that's triggered instead of the actual hit charge, and validate that the channels have the expected thresholds. This also answers the Issue #250.
image
All entries show that they are within the expected difference of the input thresholds and the threshold value after ADC quantization. The bin at -0.9 above for instance corresponds to channels with a threshold of 1 ke-, which after digitization gives a value of ~0.07 ke-.


larndsim/config/config.yaml
```diff
@@ -45,6 +45,8 @@ module0:
     PIXEL_LAYOUT_ID: [0, 0, 1, 0] 
     RESPONSE:        [response_44_v2a_50ns.npy, response_38_v2b_50ns.npy]
     RESPONSE_ID:     [0, 0, 1, 0] 
+    PIXEL_THRESHOLDS_FILE: [/global/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/2x2_CRS_thresholds/thresholds_module0.npz, /global/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/2x2_CRS_thresholds/thresholds_module1.npz, /global/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/2x2_CRS_thresholds/thresholds_module2.npz, /global/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/2x2_CRS_thresholds/thresholds_module3.npz]
```
krwood, line 48: Hardcodes a dependence on NERSC filesystems. Are these files sufficiently small to include in the `bin` directory? Otherwise maybe we could include a script that downloads these from the nersc portal into the local bin directory?

krwood (approved):
Looks good to me. I'm a little surprised at how large the threshold residuals are just from digitization effects, but I haven't done the math.

I'm okay with merging this even with the hardcoded nersc paths, but we might want to think of a better way of doing this in the future. We might include a get_inputs.sh script that is run after installation (as we do in ndlar_flow).

jaafar-chakrani:
Given the discussion in #267, I realized that this applies the thresholds in the simulation without taking into account tile swaps present in the data. I will update the input files to reflect these swaps.

jaafar-chakrani:
Files updated on NERSC with the tile swaps.

PR #266, opened 2025-02-26
correct impulse tick size

YifanC:


krwood:
FTR here is a comparison of waveforms from the same 2x2 NuMI simulated event before Yifan's fix using:

`sipm_response_model: 1`
sanity_check_lrs8ns_v1_evt1

`sipm_response_model: 0`
sanity_check_lrs8ns_v2_evt1

PR #267, opened 2025-03-05
Fix drift direction assumption

diaza:
Line 343 in the old `larndsim/consts/detector.py` determines the drift direction based on the index of the tpc. I changed it so that it uses the `tile_orientation` field to determine the drift direction (like how ndlar_flow does [here)](https://github.com/DUNE/ndlar_flow/blob/7dd1aad6981a721be3f410dc9d79d782fbbdd4aa/src/proto_nd_flow/resources/geometry.py#L745).
The above fix will allow a future correction (which is read to push if this request is approved) where the positions of the simulation io_groups will match the data's (Currently they're swapped within each module). I assume this was not possible before since the code forced a `drift_direction = 1` for io_groups [1, 3, 5, 7], when `drift_direction` should be in the negative x direction if we wanted to place them in their true position. (Note: in the code, `drift_direction` isn't really the drift direction, but the direction the charge drifts _from_. I changed the code to say "cathode direction," where its the direction of the cathode relative to the anode tile).


diaza:
I've added a new commit that includes updated tile layout files. The old tile layout files still work, but these newer ones will place the io_groups in the right position. 
The filenames should be improved, but I don't know if there's a version naming scheme? If the pull is approved, these would be the files that also go into ndlar_flow

alexbooth92 (commented):
Thank you for the tag @diaza. These new pixel layout files, do they apply only to 2x2? I guess when we are talking about io_groups being in the right place you mean with respect to how the 2x2 is actually constructed which is less of a concern for NDLAr?

diaza:
@alexbooth92 Yeah, it should ideally only affect 2x2. Just wanted you to be aware in case issues arise.

YifanC:
Hi @diaza, I think this is a useful change! Although I have one question and a comment.
Comparing pixel layout for example `multi_tile_layout-2.4.16_v3_driftfix.yaml` and `multi_tile_layout-2.4.16_v3.yaml`, I think you basically swapped tpc1 and tpc2 positions? Is it for aesthetic reasons so that data and mc io would be in the same location and the same naming convention? ([S5 of your slides](https://docs.dunescience.org/cgi-bin/private/RetrieveFile?docid=33217&filename=20250306-NDPrototypes-disabledchannels.pdf)) Given your changes on how to get the drift direct, should the simulation with `multi_tile_layout-2.4.16_v3_driftfix.yaml` or `multi_tile_layout-2.4.16_v3.yaml` yield the exact the same result?
If I understand you correctly, why not replace `multi_tile_layout-2.4.16_v3.yaml` with `multi_tile_layout-2.4.16_v3_driftfix.yaml` with the name of `multi_tile_layout-2.4.16_v4.yaml` or so?

diaza:
Hi @YifanC!
Yes, the new layout files just swap the positions of the anodes within modules so that io_groups align between data and simulation. This corrected layout didn't work before the correction in `larndsim/consts/detector.py`. 
It's partly aesthetic, but it also starts the ground work for having simulation with the same tile layout as the data. This should be less error prone when e.g. we add thresholds or gain calibrations derived from data 
When I looked at event displays, running larnd-sim and ndlar_flow with the updated layout gave identical tracks as when I ran with the old layouts. 
I can update the filenames soon

YifanC:
Sorry for my weird choice of word. I agree with the need to unify the tpc naming convention in data and simulation, thus the swaps.

YifanC:
@diaza Does it make sense to also remove v3 to avoid confusions/redundancies?

diaza:
@YifanC Well, I don't think it should be left just as `multi_tile_layout-2.4.16.yaml` since, to me, the `2.4.16` implies that this is some particular version that doesn't change. Either we remove all version numbering so that there's no implication that this is a constant file, or we update the version number. I don't know what the scheme behind `2.4.16` for Modules[0,1,3] and `2.5.16` for Module2 is though. Maybe we update 2.X.16 -> 2.X.17?

YifanC:
There is some versioning convention [here](https://github.com/larpix/larpix-geometry/blob/master/larpixgeometry/layouts/multi_tile_layout-2.4.16.py):
```
- multitile_layout_version: version of the multi-tile layout in the X.Y.Z. format
    where X represents the ASIC version, Y represents the single-tile version
    (incremented from 0), and Z represents the number of tiles in the layout.
```
Maybe Jaafar can comment.
Removing the not used the layout is just a suggestion.

YifanC (approved):

diaza:
Ah, so the "2.4.16" is more of a descriptor instead of a software version number? Ok, I think it makes sense to get rid of the "v4" soon. What would make sense would be to update the files here and in flow simultaneously, but I'm concerned that confusion might come up later if people use flow with the updated layout but on old MiniRun6 files...
I'll approve the pull request so that we have the corrected `larndsim/consts/detector.py` in place, and deal with the naming of the layout files right before we get into production.

PR #270, opened 2025-03-17
Add pixel threshold files to repository

jaafar-chakrani:
Following the suggestion in #263, I added the `.npz` input files with the pixel thresholds to the `larndsim/bin` folder. This required me to make a change in `larndsim/config/config.py` to the function that checks that the `larndsim/config/config.yaml` configuration has all the keys of `CONFIG_DIR`, by skipping this check for the pixel threshold files since they're optional. Not sure if this is the best approach, so any comments or suggestions are welcome!
@krwood @YifanC @mjkramer


krwood (approved):
Looks good to me

Issue #271, opened 2025-03-31
CUDA memory access issue


sam-fogarty:
While trying to simulate a large sample of 100k gammas, I encounter a CUDA error that looks like:

> Simulating batches...:   0%|                | 137/52228 [00:06<41:17, 21.03it/s]/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/cli/simulate_pixels.py:1123: UserWarning: 
Traceback (most recent call last):                                              
  File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/cli/simulate_pixels.py", line 1376, in <module>
    fire.Fire(run_simulation)                                                   
  File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/larndsim.venv/lib/python3.11/site-packages/fire/core.py", line 135, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/larndsim.venv/lib/python3.11/site-packages/fire/core.py", line 468, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
                                ^^^^^^^^^^^^^^^^^^^^
  File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/larndsim.venv/lib/python3.11/site-packages/fire/core.py", line 684, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^
  File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/cli/simulate_pixels.py", line 1106, in run_simulation
    pixels_tracks_signals = cp.zeros(len(detector.TIME_TICKS) * int(num_backtrack.sum()))
                                                                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "cupy/_core/core.pyx", line 1771, in cupy._core.core._ndarray_base.__int__
  File "cupy/_core/core.pyx", line 1910, in cupy._core.core._ndarray_base.get
  File "cupy/cuda/memory.pyx", line 586, in cupy.cuda.memory.MemoryPointer.copy_to_host_async
[...]

PR #274, opened 2025-04-04
Prevent out of bounds indexing error

sam-fogarty:
In `detsim.get_track_pixel_map2`, it is possible that the function will try to index into `unique_pix` with an index > unique_pix.shape[0], causing an error. To prevent this, I add a simple check if the index is >= unique_pix.shape[0], if so return. This is similar to other implementations in other functions that prevent such an indexing error.

For reference, here is an example error I get without this change:

> Simulating batches...:   0%|                | 137/52228 [00:06<41:17, 21.03it/s]/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/cli/simulate_pixels.py:1123: UserWarning: 
> Traceback (most recent call last):                                              
>   File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/cli/simulate_pixels.py", line 1376, in <module>
>     fire.Fire(run_simulation)                                                   
>   File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/larndsim.venv/lib/python3.11/site-packages/fire/core.py", line 135, in Fire
>     component_trace = _Fire(component, args, parsed_flag_args, context, name)
>                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/larndsim.venv/lib/python3.11/site-packages/fire/core.py", line 468, in _Fire
>     component, remaining_args = _CallAndUpdateTrace(
>                                 ^^^^^^^^^^^^^^^^^^^^
>   File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/larndsim.venv/lib/python3.11/site-packages/fire/core.py", line 684, in _CallAndUpdateTrace
>     component = fn(*varargs, **kwargs)
>                 ^^^^^^^^^^^^^^^^^^^^^^
>   File "/pscratch/sd/s/sfogarty/2x2_MC_job/larnd-sim/cli/simulate_pixels.py", line 1106, in run_simulation
>     pixels_tracks_signals = cp.zeros(len(detector.TIME_TICKS) * int(num_backtrack.sum()))
>                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^
>   File "cupy/_core/core.pyx", line 1771, in cupy._core.core._ndarray_base.__int__
>   File "cupy/_core/core.pyx", line 1910, in cupy._core.core._ndarray_base.get
>   File "cupy/cuda/memory.pyx", line 586, in cupy.cuda.memory.MemoryPointer.copy_to_host_async
>   File "cupy_backends/cuda/api/runtime.pyx", line 607, in cupy_backends.cuda.api.runtime.memcpyAsync
>   File "cupy_backends/cuda/api/runtime.pyx", line 146, in cupy_backends.cuda.api.runtime.check_status
> cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorIllegalAddress: an illegal memory access was encountered
> ERROR:numba.cuda.cudadrv.driver:Call to cuModuleUnload results in UNKNOWN_CUDA_ERROR

### DIFF
--- a/larndsim/detsim.py
+++ b/larndsim/detsim.py
@@ -550,7 +550,8 @@ def get_track_pixel_map(track_pixel_map, unique_pix, pixels):
     # the index is with respect to the total number segments in the batch
     # so when it is translated to "segment_id", it is correct
     index = cuda.grid(1)
-
+    if index >= unique_pix.shape[0]:
+        return
     upix = unique_pix[index]
 
     for itrk in range(pixels.shape[0]):
@@ -582,7 +583,8 @@ def get_track_pixel_map2(track_pixel_map, unique_pix, pixels, distances, max_dis
     """
     # index of unique_pix array
     index = cuda.grid(1)
-
+    if index >= unique_pix.shape[0]:
+        return
     upix = unique_pix[index]
 
     for target_dist in range(max_distance):


diaza:
(I'm not sure who the CUDA experts are, so I just added everyone as reviewers)
So the way you handle the index error is inline with other places in the code? If the treatment is consistent, then I'm fine with the current fix. But could the if statement be hiding some underlying bug elsewhere in the code?

diaza:
Resolves #271

sam-fogarty:
Hi Alex, the change makes the function match how this is handled in other functions, i.e.
https://github.com/DUNE/larnd-sim/blob/develop/larndsim/detsim.py#L116 
https://github.com/DUNE/larnd-sim/blob/develop/larndsim/detsim.py#L240

So for this reason, I conclude that the treatment is consistent across the different functions.

diaza:
Ok, I'll wait if there's input from the others. If not I'll merge tomorrow. Thanks for digging into this!

YifanC (approved):

YifanC:
@sam-fogarty The change is in `detsim.get_track_pixel_map` and your comment indicates the problem is in`detsim.get_track_pixel_map2`. I don't know if it's a typo. If you meant `detsim.get_track_pixel_map`, the change makes sense, but it's not currently used, so I'm not sure I fully understand the comment. The PR itself is fine

sam-fogarty:
@YifanC The change is made in both `detsim.get_track_pixel_map` and `detsim.get_track_pixel_map2`. Only the latter is used in the simulation right now which is where the error arrises, but I made the change in both functions for completeness (as the error would happen for both functions).

YifanC:
> @YifanC The change is made in both `detsim.get_track_pixel_map` and `detsim.get_track_pixel_map2`. Only the latter is used in the simulation right now which is where the error arrises, but I made the change in both functions for completeness (as the error would happen for both functions).

Yea, you are right. The check is add to both `detsim.get_track_pixel_map` and `detsim.get_track_pixel_map2`. Sorry I misread. However, I don't understand how you would have index out of bound. The kernel size is set by the `unique_pix`. Could you elaborate how did you get to that error? 

Again, regardless the question above, it doesn't affect I think the PR can get into the develop

sam-fogarty:
I don't fully understand why, but it just seems to happen. I reckon it is why this line is used in other functions. I do think it would be good to understand why this happens. But I would have to ask for the feedback of someone more well-versed in CUDA than me. But I don't think this is a barrier to merging this to develop.

PR #275, opened 2025-05-16
FSD light noise update

cuddandr:
Includes updates to the yamls to use a new light noise file for the FSD simulation.

This also includes a couple changes to `simulate_pixels.py` to account for inaccuracies in floating point math when calculating `digit_samples`, and a change to the logic statement for loading the light LUT.

The new calculation of `digit_samples` has been checked using the 2x2 and ND-LAr values to give the same result.


larndsim/detector_properties/fsd.yaml
```diff
@@ -32,11 +32,12 @@ tau_t: 1.530 # us
...
 op_channel_efficiency: 1
 light_gain: -7.0
+light_tick_size: 0.016 # us
 light_det_noise_sample_spacing: 0.016 # us
 light_trig_threshold: [-2000, -4500] # [LCM, ACL]
 light_trig_mode: 1
-light_window: [0, 16] # us
-light_trig_window: [1.6, 14.4] # us
+light_window: [0.00, 8.64] # us
+light_trig_window: [0.96, 8.64] # us
```
krwood, line 41: I _think_ this should be

`light_trig_window: [0.96, 7.68] # us`

But should be double checked with an expert...

cuddandr: @AWh1t3 and I worked out the `light_trig_window` (and `light_window`) to get the correctly sized FSD waveforms, but I'll have to ask again why this is the correct window (namely the end point being the same as `light_window`).

PR #276, opened 2025-06-12
Feature pedestals

diaza:
This pull request implements pedestal variations into the simulation. The pedestal variations are implemented in a way very similar to the gain. 

Before going into the pedestal variations, I'd like to discuss a necessary change I made which I'd like to alert others to in case it might have unintended consequences.
For context:
The digitization of the packets happen [here](https://github.com/DUNE/larnd-sim/blob/77a49e9a208a51d43df6057354109c6e70d5490d/cli/simulate_pixels.py#L1213)
`adc_list = fee.digitize(integral_list, gain_list, pedestal_list)`,
where `integral_list` is a 2D array where the first axis has length equal to the number of unique pixels being digitized, and the second axis is each packet in accumulated voltage. The length of the second array is fixed to `sim.MAX_ADC_VALUES` (30, in our case). The vast majority of elements are `0`. 
In the function [digitize](https://github.com/DUNE/larnd-sim/blob/37db5293f5a2d92421046b88228ca3f97380cb7d/larndsim/fee.py#L514), even if an element in `integral_list` is 0 (and therefore no trigger), the function returns a non-zero value because of the pedestal. So `integral_list` is filled with non-zero adc values. Then, in the function [export_to_hdf5](https://github.com/DUNE/larnd-sim/blob/37db5293f5a2d92421046b88228ca3f97380cb7d/larndsim/fee.py#L84C5-L84C19), there's a [line](https://github.com/DUNE/larnd-sim/blob/37db5293f5a2d92421046b88228ca3f97380cb7d/larndsim/fee.py#L177)
`if adc > digitize(0):`
My guess is that this line is meant to filter out the ADC elements that actually had no hits in `integral_list`. The problem with this line is that it uses default pedestals, so it would have to be modified for the pedestal variations. 
To resolve this, I instead added the following [line](https://github.com/DUNE/larnd-sim/blob/77a49e9a208a51d43df6057354109c6e70d5490d/larndsim/fee.py#L530C1-L530C33) to `digitize` 
`adcs[integral_list == 0] = 0`
and changed the above line in `export_to_hdf5` [to](https://github.com/DUNE/larnd-sim/blob/77a49e9a208a51d43df6057354109c6e70d5490d/larndsim/fee.py#L177) 
`if adc > 0:`. This avoids the earlier issue of having varied pedestals, and I think it makes more sense to have the `adcs` variable to have empty entries where there were no triggers. 
This slightly changes the number of outputted packets. In principle, there can be triggers where the truncated adc output is still equal to `digitize(0)`, and were thrown out. This also stopped further hits from the same pixel from being saved, since `export_to_hdf5` had a `break` if `adc > digitize(0)` was not satisfied.
Below is the result of some tests I did. I ported in the code from the feature_reproducibility branch for this comparison, and left out pedestal variations. First, I checked that all packets that were outputted before the above edits were also outputted after the edits. Second, I output all the hits that are new with the above edits. You can see in the output below that all of the new hits are composed of sequences of hits from the same pixels, where the first one in each sequence has the adc value of 23 (the default pedestal). 
image

To test the pedestal implementation, I again port over the reproducibility code and compared hits with default pedestals and with varied pedestals. I take the difference of the ADC values for each hit with and without pedestal variation. Given [this](https://github.com/DUNE/larnd-sim/blob/77a49e9a208a51d43df6057354109c6e70d5490d/larndsim/fee.py#L527C67-L527C75) equation, that value should be consistent with `(pedestal_from_file - 580.)*256/(v_ref-v_cm)`. Where pedestal_from_file is the pedestal extracted from the inputted pedestal file for that given pixel, and 580 is the default pedestal voltage. Below is the code used, and the outputs
```python
with open('2x2_sim/run-ndlar-flow/ndlar_flow/data/proto_nd_flow/reference-cold-pedestal-2024_06_05_08_28_19_CDTevd_ped.tile_id_swapped.decimal.json', 'r') as data_file:
    data_swapped = json.load(data_file)

default = f2['packets']['io_group','io_channel', 'chip_id', 'channel_id', 'dataword'][mask2].astype(np.dtype([('io_group', 'int64'), ('io_channel', 'int64'), ('chip_id', 'int64'), ('channel_id', 'int64'), ('dataword', 'int64')]))
default_tile_id = ((default['io_channel']) - 1) //4 +1 + (default['io_group']-1)*8
default_unique_id = ((default['io_group']*10000+default_tile_id)*1000 + default['chip_id'])*100 + default['channel_id']

varied = f3['packets']['io_group','io_channel', 'chip_id', 'channel_id', 'dataword'][mask3].astype(np.dtype([('io_group', 'int64'), ('io_channel', 'int64'), ('chip_id', 'int64'), ('channel_id', 'int64'), ('dataword', 'int64')]))
varied_tile_id = ((varied['io_channel']) - 1) //4 +1 + (varied['io_group']-1)*8
varied_unique_id = ((varied['io_group']*10000+varied_tile_id)*1000 + varied['chip_id'])*100 + varied['channel_id']

n=len(varied)
mv_diff = np.zeros(n)
adc_diff = np.zeros(n)
adc_counts = 256
v_ref = 1568.
v_cm=478.1
n_notfound=0
n_maxed = 0
for i in range(n):

    if not varied['dataword'][i]<255:
        n_maxed+=1
        continue
    try:
        ped = data_swapped[str(varied_unique_id[i])]
        mv_diff[i] = (ped['pedestal_mv'] - 580.)*256/(v_ref-v_cm)
        
    except KeyError:
        n_notfound+=1
        mv_diff[i] = 0
        
    adc_diff[i] = varied['dataword'][i] - default['dataword'][i]

print("Number of total hits: ", n)
print("Number of hits without pedestal in file: ", n_notfound)
print("Number of satured hits: ", n_maxed)

print("Are all extracted adc differences within expectation?") 
print(np.all( np.abs(mv_diff - adc_diff) < 1))
```
```
Number of total hits:  67987
Number of hits without pedestal in file:  333
Number of satured hits:  120
Are all extracted adc differences within expectation?
True
```


diaza:
This pull request should be merged in conjunction with the [pull request](https://github.com/DUNE/ndlar_flow/pull/197) in flow.

diaza:
Hmm, I'm having some issues running larnd-sim with light. Might be because I'm on an older commit of `develop`. I need to check, so I'll convert this back to a draft in the meantime.

diaza:
Ok, seems to work fine with the latest from `develop` merged in.

diaza:
Hi friends, if there's no comments on this by ~Friday I'll merge it into develop.

PR #277, opened 2025-06-17
Don't generate ionization electrons and scint for gamma/neutron segments

wforeman:
Simple check in `quenching.py` to exclude common neutral particles like gammas and neutrons (with non-zero 'dE') from the charge/light generation process. It's not 100% clear why this happens, but the leading theory is that the total 'dE' returned by edepsim also includes non-ionizing energy loss. A more comprehensive solution would be to find a way to reliably calculate only the ionizing dE and use that in the calculation of `n_electrons` and `n_photons` instead of total dE. This would require upstream changes to the ROOT-to-HDF5 conversion stage.

James Mead ran some code over a 2x2 simulation test file before and after this fix. Before:
image (3)

After:
image (4)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ docs/build
 .vscode
 examples/module0_corsika.h5
 test
+*.*.swp
--- a/larndsim/quenching.py
+++ b/larndsim/quenching.py
@@ -25,20 +25,29 @@ def quench(tracks, mode):
     if itrk < tracks.shape[0]:
         dEdx = tracks[itrk]["dEdx"]
         dE = tracks[itrk]["dE"]
+        # TODO: ensure 'dE' is visible energy (may require subtracting off the SecondaryEnergyDeposit
+        #       calculated by edepsim that accounts for non-ionizing energy loss)
+        
+        # initialize to zero
+        tracks[itrk]["n_electrons"] = 0
+        tracks[itrk]["n_photons"]   = 0
 
-        recomb = 0
-        if mode == physics.BOX:
-            # Baller, 2013 JINST 8 P08005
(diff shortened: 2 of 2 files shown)


sam-fogarty:
I see no issues with these changes.

krwood:
I'm not convinced this change is correct. I think we need to verify what edep-sim is filling into `dE` before doing this. For example, what happens when we get neutron signals from capture? I was under the impression it would create a segment with dE > 0 and a neutron PDG code.

krwood:
Okay, that doesn't seem to be the case. At least in some files from a simulation of the neutron sources we are planning to run with at 2x2.
Screenshot from 2025-06-25 12-04-09

wforeman:
@krwood  is this a n-capture producing a gamma that pair-produces into an e+e- pair?

I would expect a neutron capture to produce gammas which would then create electrons via compton scattering or pair production. It should be the electron segments that produce the charge and light, not the neutron or its gammas. In either case, if it's the neutron or gamma segment that's used, the dE/dx (used to calculate recombination in the charge/light generation) would not make sense since it's not actually ionizing anything.

krwood:
@wforeman I totally agree with you from the standpoint of the physics. What I am unsure about is the bookkeeping.

wforeman:
I'm unsure too. Maybe these neutron energy losses are from very low-energy elastic scatters (< 0.1 MeV) that occur along the trajectory as they thermalize, in which case we don't want to turn them into signals. Ideally any charged particle high enough in energy to generate charge/light will be tracked by Geant and be saved into separate edepsim segments. Needs further study, but I think this safeguard for now is reasonable.

PR #278, opened 2025-06-18
Include event ID in mc_packet_assn for ALL packets

mjkramer:
Previously it was being set to -1 for non-data packets. This change makes it easier to reorder the packets event-by-event (to make life easier for flow's event builder).

Tested and confirmed to fill in the right event ID for all packets.


mjkramer:
resubmitting as a clean PR

Issue #125, opened 2023-03-31
Reproducibility


krwood:
It seems that larnd-sim is not creating a reproducible simulation - at least when running our 2x2 spill edep-sim files through. When running the same file through simulate_pixels.py with the random seed fixed to the same value, I get a different number of packets generated for each. This is true even when disabling the light simulation. (tests run on feature_config_seed branch)

Produce the two simulation files:
```
#!/bin/bash

EDEPH5="/global/cfs/cdirs/dune/www/data/2x2/simulation/mkramer_dev/MiniRun2_1E19_RHC_convert2h5/EDEPSIM_H5/g4numiv6_minervame_me000z-200i_0_0001.000.EDEPSIM.h5"
LARNDH5="g4numiv6_minervame_me000z-200i_0_0001.000.EDEPSIM.LARNDSIM-TEST"
ROOT_LARNDSIM="/global/cfs/cdirs/dune/users/kwood/larnd-sim_reprod/larnd-sim"

if test -f "$LARNDH5"; then
    rm $LARNDH5
fi 

python simulate_pixels.py --input_filename=${EDEPH5} \
  --output_filename=${LARNDH5}"_1.h5" \
  --detector_properties=${ROOT_LARNDSIM}/larndsim/detector_properties/2x2.yaml \
  --simulation_properties=${ROOT_LARNDSIM}/larndsim/simulation_properties/2x2_NuMI_sim.yaml \
  --pixel_layout=${ROOT_LARNDSIM}/larndsim/pixel_layouts/multi_tile_layout-2.3.16.yaml \
  --response_file=${ROOT_LARNDSIM}/larndsim/bin/response_44.npy \
  --light_det_noise_filename=${ROOT_LARNDSIM}/larndsim/bin/light_noise-2x2-example.npy \
  --rand_seed=12345

python simulate_pixels.py --input_filename=${EDEPH5} \
  --output_filename=${LARNDH5}"_2.h5" \
  --detector_properties=${ROOT_LARNDSIM}/larndsim/detector_properties/2x2.yaml \
  --simulation_properties=${ROOT_LARNDSIM}/larndsim/simulation_properties/2x2_NuMI_sim.yaml \
  --pixel_layout=${ROOT_LARNDSIM}/larndsim/pixel_layouts/multi_tile_layout-2.3.16.yaml \
  --response_file=${ROOT_LARNDSIM}/larndsim/bin/response_44.npy \
[...]

PR #279, opened 2025-06-27
Feature reproducibility

diaza:
A work in progress for adding reproducibility to larnd-sim. Looks to resolve issue #125.

Current status: reproducibility seems to be there. Need to look at the flow outputs and see if it's sensible. Need to check if time consumption increases appreciably. Maybe other improvements? 

In the develop branch:
The `signals` [variable](https://github.com/DUNE/larnd-sim/blob/ad3fe922768b185600c6bc4cb4f6dd28d807c308/cli/simulate_pixels.py#L1063) contains the output of the induced signal on each relevant pixel from each segment. It has a shape of (# of segments x # of neighboring pixels x # time ticks). 
When [filling](https://github.com/DUNE/larnd-sim/blob/ad3fe922768b185600c6bc4cb4f6dd28d807c308/cli/simulate_pixels.py#L1072) in `signals` in `tracks_current_mc`, a thread is created for each element in the `signals` array and each thread samples from an RNG in the `rng_states` array.  In the current implementation it looks like, for each element in the first and second dimension, all the threads in the third dimension [samples from the same rng state](https://github.com/DUNE/larnd-sim/blob/ad3fe922768b185600c6bc4cb4f6dd28d807c308/larndsim/detsim.py#L169). This would lead to non-reproducibility. 

My branch:
I first change the `tracks_current_mc` function so that each [thread samples from a different rng state](https://github.com/DUNE/larnd-sim/blob/88497ce1b4fa4d48af50e107da139c7fb8b1cfcc/larndsim/detsim.py#L174). This seemed to result in reproducibility.  But this leads to another issue where, when creating enough rng states for an event with a very large signals.size, the job would run out of memory and crash. This first change is contained in this [commit](https://github.com/DUNE/larnd-sim/commit/0279e002f468eaf1e7f401ac9f78414aa98d57f2). 
My solution is that when we reach an event where more rng_states are needed to accommodate `signals`, `signals` is instead [batched](https://github.com/DUNE/larnd-sim/blob/88497ce1b4fa4d48af50e107da139c7fb8b1cfcc/cli/simulate_pixels.py#L1077) so that each batch is accommodated within the existing number of rng_states. 
Initial checks seem to give reproducibility, with no appreciable increase in run time. 

There are probably improvements that can be made. The batch sizes are determined by the size of `rng_states`, whose size is variable (but made to only grow) and changes based on other lines in the code. Maybe an optimal number can be determined before hand. 
Also, maybe the batching can be expanded to surround other steps in the code? Or maybe batching `tracks_current_mc` is enough to solve Matt's goal of reducing memory usage? Or maybe replacing `signals` with an awkward array removes all need for batching?

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -2,7 +2,7 @@
 """
 Command-line interface to larnd-sim module.
 """
-from math import ceil
+from math import ceil, floor
 from time import time
 import warnings
 from collections import defaultdict
@@ -1068,8 +1068,20 @@ def run_simulation(input_filename,
                 BPG_Y = max(ceil(signals.shape[1] / TPB[1]),1)
                 BPG_Z = max(ceil(signals.shape[2] / TPB[2]),1)
                 BPG = (BPG_X, BPG_Y, BPG_Z)
-                rng_states = maybe_create_rng_states(int(np.prod(TPB) * np.prod(BPG)), seed=rand_seed+ievd+itrk, rng_states=rng_states)
-                detsim.tracks_current_mc[BPG,TPB](signals, neighboring_pixels, selected_tracks, response, rng_states)
+
+                BPG_X_subbatch = min(floor(len(rng_states) / (np.prod(TPB) * BPG[1] * BPG[2])), BPG_X)
+                BPG_subbatch = (BPG_X_subbatch, BPG_Y, BPG_Z)
+                subbatches_size = BPG_X_subbatch
+                n_subbatches = ceil(BPG_X/subbatches_size)
+
+                for i_subbatch in range(n_subbatches):
+                    start = i_subbatch*subbatches_size
+                    end = (i_subbatch+1)*subbatches_size
+                    detsim.tracks_current_mc[BPG_subbatch,TPB](signals[start:end],
+                                                                 neighboring_pixels[start:end],
+                                                                 selected_tracks[start:end],
+                                                                 response, rng_states)
+
                 RangePop()
 
                 RangePush("pixel_index_map")
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -98,7 +98,7 @@ def maybe_create_rng_states(n, seed=0, rng_states=None):
     if n > len(rng_states):
         new_states = device_array(n, dtype=rng_states.dtype)
         new_states[:len(rng_states)] = rng_states
-        new_states[len(rng_states):] = create_xoroshiro128p_states(n - len(rng_states), seed=seed)
+        new_states[len(rng_states):] = create_xoroshiro128p_states(n - len(rng_states), seed=seed, subsequence_start = len(rng_states))
         return new_states
 
     return rng_states
@@ -1159,7 +1159,7 @@ def run_simulation(input_filename,
                 # TPB = 128
                 TPB = 4 #[1, 4, 8, 16, 32, 64, 128, 256]
                 BPG = ceil(pixels_signals.shape[0] / TPB)
-                rng_states = maybe_create_rng_states(int(TPB * BPG), seed=rand_seed+ievd+itrk, rng_states=rng_states)
+                rng_states = maybe_create_rng_states(int(TPB * BPG), seed=rand_seed, rng_states=rng_states)
                 TPB_lut = 128 # supposed to be 128
                 BPG_lut = ceil(pixels_signals.shape[0] / TPB_lut)  
                 if pixel_thresholds_file is not None:
@@ -1252,7 +1252,7 @@ def run_simulation(input_filename,
 
                     light_sample_inc_disc = cp.zeros_like(light_sample_inc)
                     rng_states = maybe_create_rng_states(int(np.prod(TPB) * np.prod(BPG)),
-                                                         seed=rand_seed+ievd+itrk, rng_states=rng_states)
+                                                         seed=rand_seed, rng_states=rng_states)
                     light_sim.calc_stat_fluctuations[BPG, TPB](light_sample_inc_scint, light_sample_inc_disc, rng_states)
                     RangePop()
 
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -1068,7 +1068,7 @@ def run_simulation(input_filename,
                 BPG_Y = max(ceil(signals.shape[1] / TPB[1]),1)
                 BPG_Z = max(ceil(signals.shape[2] / TPB[2]),1)
                 BPG = (BPG_X, BPG_Y, BPG_Z)
-                rng_states = maybe_create_rng_states(int(np.prod(TPB[:2]) * np.prod(BPG[:2])), seed=rand_seed+ievd+itrk, rng_states=rng_states)
+                rng_states = maybe_create_rng_states(int(np.prod(TPB) * np.prod(BPG)), seed=rand_seed+ievd+itrk, rng_states=rng_states)
                 detsim.tracks_current_mc[BPG,TPB](signals, neighboring_pixels, selected_tracks, response, rng_states)
                 RangePop()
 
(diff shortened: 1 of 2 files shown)


diaza:
I've done some checks and it looks to me like I'm getting identical result. I've also attached some plots comparing Q distributions before and after my implementation, to show that there's nothing crazy going on. 
Screenshot 2025-06-30 at 3 07 12 PM

diaza:
I'm thinking about improvements to this PR, and batching in general. My musings:
There's this batching that happens [here](https://github.com/DUNE/larnd-sim/blob/88497ce1b4fa4d48af50e107da139c7fb8b1cfcc/cli/simulate_pixels.py#L918) that seem's unideal. Within each event, the batches split the event by the number of segments (I'm going to call this batching step as segment-batching, to avoid confusing with the sub-batching done for reproducibility). The different segment-batches are simulated completely separately, including the ADC triggering. If possible, I think we'd want to separate the charge segment-batching and light segment-batching, since the ideal batch sizes are probably different for the two steps. Once charge and light are split up, I think we'd want to do one of two things to the charge segment-batching
1. If the sub-batching implemented in this PR for reproducibility also solves the memory issue, then we can completely remove the charge segment-batching (or, at least, make the segment-batching size very very large)
2. Otherwise, maybe we can try pulling the steps starting at [L1172](https://github.com/DUNE/larnd-sim/blob/88497ce1b4fa4d48af50e107da139c7fb8b1cfcc/cli/simulate_pixels.py#L1172) 
```python
                fee.get_adc_values[BPG, TPB](pixels_signals,
                                             pixels_tracks_signals,
                                             num_backtrack,
                                             offset_backtrack,
                                             time_ticks,
                                             integral_list,
                                             adc_ticks_list,
                                             0,
                                             rng_states,
                                             current_fractions,
                                             pixel_thresholds)
```
outside of the `for` loop of the segment-batching. Within the `for` loop, we can append  to `pixels_signals` (and whichever other variables) for the whole event, and then pass that to `fee.get_adc_values`. This way, we can make the segment-batching arbitrarily small. 
@mjkramer is probably most familiar with possible batching issues, so I'll message him privately and discuss strategies.

diaza:
Regardless of the above, I think this PR can still be merged. There is one unrelated detail I want to check, and that's that I'm not sure if the seed-ing done when growing `rng_states` is prone to repeats. But that's unrelated to the overall PR.

diaza:
The commit above avoids repeating seeds in the rng creator. When [maybe_create_rng_states](https://github.com/DUNE/larnd-sim/blob/95ba8bc4bd02146a887554ef3d38963a8b836d87/cli/simulate_pixels.py#L92) is called, it either creates the `rng_states` array with size `n` (as in L416)
`L416 rng_states = maybe_create_rng_states(1024*256, seed=rand_seed)`
or appends to `rng_states` if the requested `n` is bigger than `len(rng_states)`, (as in L1162 and L1254). `rand_seed` is in practice the file index. 
```python
1162 rng_states = maybe_create_rng_states(int(TPB * BPG), seed=rand_seed+ievd+itrk, rng_states=rng_states)
```
```python
1254 rng_states = maybe_create_rng_states(int(np.prod(TPB) * np.prod(BPG)), seed=rand_seed+ievd+itrk, rng_states=rng_states)
```
The function would append by creating a new array with length `n`, copying the original `rng_states` to the first `len(rng_states)` elements, and creating a new rng array of length `n - len(rng_states)` into the remaining elements. 

I think that the original implementation of `seed=rand_seed+ievd+itrk` when growing `rng_states` was prone to creating identical rng_state arrays, either within the same file or between files. 

In my commit, I've modified `maybe_create_rng_states` so that when expanding `rng_states` to `n` elements, the additional elements would be in a state as if the whole `rng_states` was originally created with `n`. 

For example, using the previous commit, I print the states if (top image) I created a rng_states with originally 5 elements, and (middle image) if I originally created with 3 elements and then expanded to 5. 
Screenshot 2025-07-03 at 2 12 54 PM
Screenshot 2025-07-03 at 2 14 15 PM

With the new commit, I now create rng_states with 3 elements and then expand to 5. 
Screenshot 2025-07-03 at 2 14 19 PM
As you can see, the states are identical as if `rng_states` was created with the original 5. 

Therefore, for all calls to `maybe_create_rng_states`, the seed can be kept to `seed=rand_seed`, ensuring that there are no repeated seeds within and between files.

diaza:
Regarding this PR, I'll explore ways to make the batching better. Matt and I both think the second option [here](https://github.com/DUNE/larnd-sim/pull/279#issuecomment-3021316176) is a good place to start. But, even if there are no updates, I'm thinking I'll merge this PR ~Wednesday next week if there are no further comments.

YifanC:
As the PR is to address the reproducibility issue, would it make sense to make a bit precision comparison for the output to illustrate it does achieve the reproducibility. Logically I think it all makes sense and the rng state also looks good, but since this is a rabbit hole of other issues as well, maybe it is worth to double check the output.

YifanC:
For the discussion about `fee.get_adc_values()`, I think it makes sense. One challenge I can think of is that it will make backtracking slightly complicated, as the ranking of the contribution among segments to a pixel is done in the segment batch, so in a busy environment, decide which segments to keep for the backtracking information may require passing additional information.
The pixels_signals would have different length in the time axis but it's easy to pad.

YifanC:
As stated in PR#281 bullet point 4, once the seed issue is fixed. One may consider to reduce the `MIN_STEP_SIZE` to simulate diffusion better.

YifanC:
Looking at this `xoroshiro128p_normal_float32(rng_states, itrk * npix * nt + ipix * nt + it )`, I think it should also include `i_step` in, otherwise it doesnt match the point of chopping the segments to simulate diffusion if they all get the same index?

diaza:
> As the PR is to address the reproducibility issue, would it make sense to make a bit precision comparison for the output to illustrate it does achieve the reproducibility. Logically I think it all makes sense and the rng state also looks good, but since this is a rabbit hole of other issues as well, maybe it is worth to double check the output.

You mean doing a bit-wise comparison with e.g. `cmp`? There are some minor differences within the files that make that difficult (explained below), so I use `h5diff` to do a comparison. 
The easy to explain differences are the attributes `f['/_header'].attrs['created']` and `f['/_header'].attrs['modified']`, which are expected. 
The only other differences are found in `f['mc_hdr']`. It has to do with this [line](https://github.com/DUNE/larnd-sim/blob/0e211dc6316e078e5fb20d017246870a3a2dfda8/cli/simulate_pixels.py#L690). For reasons I don't entirely understand, that line returns a list with empty fields 
```
[('event_id', '<u4'),
 ('', '|V4'),
 ('vertex_id', '<u8'),
 ('vertex', '<f8', (4,)),
 ('target', '<u4'),
 ('reaction', '<i4'),
 ('isCC', '|b1'),
 ('isQES', '|b1'),
 ('isMEC', '|b1'),
 ('isRES', '|b1'),
 ('isDIS', '|b1'),
 ('isCOH', '|b1'),
 ('', '|V2'),
 ('Enu', '<f4'),
 ('nu_4mom', '<f4', (4,)),
 ('nu_pdg', '<i4'),
 ('Elep', '<f4'),
 ('lep_mom', '<f4'),
 ('lep_ang', '<f4'),
 ('lep_pdg', '<i4'),
 ('q0', '<f4'),
 ('q3', '<f4'),
 ('Q2', '<f4'),
 ('x', '<f4'),
 ('y', '<f4'),
 ('', '|V4')]
```
which then leads to extra fields in larnd-sim's output of `mc_hdr` that are seemingly filled with junk. When I do a comparison of `f['mc_hdr']` excluding those junk fields, they are identical. 

> Looking at this xoroshiro128p_normal_float32(rng_states, itrk * npix * nt + ipix * nt + it ), I think it should also include i_step in, otherwise it doesnt match the point of chopping the segments to simulate diffusion if they all get the same index?

This is not an issue. When `xoroshiro128p_normal_float32` is called, it also updates the RNG state at the requested index.

diaza:
Since the discussion regarding `fee.get_adc_values` can be separated from the reproducibility fix, I think I'll work on that in a different PR and merge this one within the next ~1-2 day, unless there are further comments.

PR #281, opened 2025-07-06
Small fixes

YifanC:
The first and second bullets of the following list described in this PR are to address the induction signal coverage problem Yousen pointed out in [this set of slides](https://indico.fnal.gov/event/69918/contributions/317688/attachments/189254/261292/PixelReadoutMeeting20250626v2.pdf).

1. The radius for the neighbouring pixels of a segment is set to 4 to match with the range of the charge response LUT, where previous it can only be 1.
2. The signal waveform range now consider the possible diffusion range to 5 sigmas. (This covers PR#280.)
3. Remove the t0 part from `signals` which previously caused trouble when late segments from decay coming in the scope as pointed out by @wforeman. I moved the segments t0 filter to a different place so it does not only apply to beam simulation. The t0 filter is further tightened with the assumption the late signals are very small to generate hits by themselves, so if there's no overlap to the main signal it can be discarded. I set this value to be 30us, which in my opinion is conservative. It can be configured in the yaml if other values are considered better.
4. Adjusted the `MIN_STEP_SIZE` which controls the segment chopping the current calculation stage. I previously increased it to roughly match the response bin size, but later realised that it should be finer to represent the diffusion distribution. (0.0064cm -> 0.001 cm) ~~Changed it back to 0.0064cm. Before the seed for diffusion is fixed (the problem and the solution discussed in PR#279), there is no point to reduce is the step size.~~ (It was my misunderstanding)
5. Fix `overlapping_segment()` so it no longer gets rid of segments that go vertically into the pixel plane.
6.  The response covers a 9x9 region, but previously the pixel signals and backtracking previously discard the corners outside of the blue square given a design choice. In this PR, the full response 9x9 regions are considered in the signal summation and backtracking while retaining the distance ranking. Note that this distance ranking is from the neighbouring pixels to the center pixels, not to the segment centers. There might be edge cases that it is not the best representations.

sketch 001

The test uses two single segments with the same length, drift position and energy deposition, which locates right above the center of the pixels 55193 and 55196 (3 pixels apart) in the module0 setup, vertically into the pixel plane. You can see the signal generated on pixel 54629 which is 4 y and z pixels away from 55193, as well as the impact of segment 1 (above 55196) on the signal generated on 55193. The induction signals on 55193 and 54629 from segment 1 and 0 are very small, but it also depends on the energy deposition. The test segment has a dE/dx of 2MeV/cm.

PR illustration 001
PR illustration 002
PR illustration 003

I tested with MiniRun5 file123 as an input. It chokes on memory with module1 event 181 by setting up signals with size (n_segments, max_number_of_neighbouring_pixels, time ticks) with or without light simulation. The neighbouring pixels include the active center pixels. The changes mainly comes from 1., but 6. in general also increases memory usage. Therefore, this is a draft PR.
Three comments on this matter:
- We can remove the edits of 6, but probably should remove the upstream calculation as well so to reduce the inflation from 1.
- We can make signals a ragged array, and make it (segments x neighbouring pixels, time ticks).
- The above two patches are relatively easy to achieve, and can solve the problem in front of us, if we insist to have all 4 neighbouring pixels enabled. To test the fidelity of the simulation, one can consider using cosmic data, which could somehow bypass this large memory use issue. However, the current simulation with segment-based logic is not very scalable (as discussed before), which is also relevant for what @diaza mentioned in PR#279. Considering the effort patching, it may not be crazy to think of a "point-charge" based simulation which may simplify the simulation memory usage and backtracking. The light part would be fairly similar. As far as I know, there are a few implementation in JAX, PyTorch serving this purpose which could be eventually considered.


yszhang95:
Hi @YifanC ,

Thank you for preparing the PR. I have a few thoughts.

--- Fix in hit level.

Bullet 6 deals with signal summation on the pixel pad and backtracking. It walks the track history to identify which contributions belong to each pixel. The key issue is the presence of a spatial window that restricts how far back we track. As long as this window remains, the fix is only partial, and the underlying inaccuracy in hit attribution persists. To fully resolve the issue, the window must be removed entirely. With that in mind, it becomes less motivated to retain the changes in bullet 1 while removing bullet 6.

On the other hand, if memory usage remains a concern and there’s a possibility we may not keep the edits in bullet 6, it may make sense to reconsider the extent of the changes in bullet 1 as well. In particular, we could explore aligning the neighboring pixel coverage in bullet 1 with the spatial window still present in bullet 6. This would maintain consistency between signal summation and backtracking, and help contain memory growth.


--- Memory consideration
I agree that keeping track of induced current in particle level is not scalable. We are developing [TRED](https://github.com/yszhang95/tred/tree/master) (written in pytorch) to address the problem. The back tracking is under development but we hope we will finish soon.

--- Decision in backtracking of truth

Back-tracking in the existing algorithm can generate misleading “positive” hits, since only the positive portions of a bipolar signal can trigger a hit—even when no charge is actually collected on that pixel. How we choose to record truth information should depend on downstream needs. For example, reconstruction teams may prefer to see only the genuinely collected (“true”) positive hits, rather than these “false” positives.

I'm still uncertain about the best path forward for backtracking. There are a few options, each with trade-offs:
- No changes: This avoids additional memory overhead and keeps the current backtracking logic unchanged. Whether far-field triggering plays a significant role in data is still under investigation. Expanding the response window to include far-field effects would significantly increase memory usage. Notably, the current algorithm—which omits far-field triggering—produces more accurate estimates of true ionization charge, as it avoids labeling small induced signals as true hits (page 8 and 9 in [slides](https://indico.fnal.gov/event/69918/contributions/317688/attachments/189254/261292/PixelReadoutMeeting20250626v2.pdf)). 
- Add far-field effects, but keep current backtracking logic (as in this draft PR): This extends the physics modeling, but the backtracking will start labeling small far-field induced charges as truth, potentially overestimating true ionization. It’s not clear yet how this would interact with calibration or downstream corrections.
- Add far-field effects and mark induced false-positive charges explicitly: This would require a redesign in larnd-sim to distinguish between direct and long-range induction (non-ionization) contributions. A signal-processing or unfolding layer would then be responsible for recovering the true ionization charge from the reconstructed hits. This may be similar to 'deghosting' step in SPINE.
- Add far-field effects and implement a new signal-processing chain: This could follow the direction of Sergey’s toy studies ([link](https://indico.fnal.gov/event/68916/contributions/316050/attachments/187968/259272/PixSi_Intro_Martynenko.pdf)), where a dedicated chain is designed to handle and disentangle far-field effects in a physically motivated way.

--- Some minor comments.

Default setup of MAX_RADIUS  for field response, and the radius for finding overlapped track segments, may failed to handle the diffusion from track segments.

larndsim/detsim.py
```diff
@@ -300,35 +317,30 @@ def get_track_pixel_map(track_pixel_map, unique_pix, pixels):
...
         for itrk in range(pixels.shape[0]):
 
             for ipix in range(pixels.shape[1]):
                 pID  = pixels[itrk][ipix]
                 dist = distances[itrk][ipix]
 
                 if (upix == pID):
-                    if (dist == target_dist): 
-
+                    if (dist == target_dist):
```
yszhang95: Comparing two floats may not be guaranteed.

larndsim/detsim.py
```diff
@@ -189,17 +202,18 @@ def tracks_current_mc(signals, pixels, tracks, response, rng_states):
                     continue
                 if y_dist > detector.RESPONSE_BIN_SIZE * response.shape[1]:
                     continue
-                if (this_time - t['t0'] + shift_t_collect) < 0 or (this_time - t['t0'] + shift_t_collect) > detector.RESPONSE_MAX_TIME:
+                if (this_time + shift_t_collect) < 0 or (this_time + shift_t_collect) > detector.RESPONSE_MAX_TIME:
```
yszhang95, line 204: The diffusion should also be included here otherwise it will be cutoff when ionization is close to anode.

larndsim/detsim.py
```diff
@@ -300,35 +317,30 @@ def get_track_pixel_map(track_pixel_map, unique_pix, pixels):
...
     """
     # index of unique_pix array
     index = cuda.grid(1)
     if index >= unique_pix.shape[0]:
         return
     upix = unique_pix[index]
 
-    for target_dist in range(max_distance):
-    
+    for target_dist in detector.NEIGHBORING_PIX_DIST:
```
yszhang95, line 333: The link from track to pixels is subject to `MAX_RADIUS`.

The link from pixel to tracks is subject to `NEIGHBORING_PIX_DIST`.  It may be error-prone to have two parameters. Enlarging `NEIGHBORING_PIX_DIST` may not work when it is larger than `MAX_RADIUS` because the map from pixel to tracks require information from the link from track to pixels.

larndsim/pixels_from_track.py
```diff
@@ -227,8 +221,8 @@ def get_neighboring_pixels(active_pixels, radius, neighboring_pixels, neighborin
         if active_pixels[pix] == -1:
             continue
 
-        for x_r in range(-radius, radius+1):
-            for y_r in range(-radius, radius+1):
+        for x_r in range(-detector.MAX_RADIUS, detector.MAX_RADIUS+1):
+            for y_r in range(-detector.MAX_RADIUS, detector.MAX_RADIUS+1):
```
yszhang95, line 225: Same comment as in `get_track_pixel_map2`.

larndsim/consts/detector.py
```diff
@@ -396,6 +383,39 @@ def set_detector_properties(detprop_file, pixel_file, response_file=None, i_modu
...
+            if isinstance(response_file, list):
+                if i_module < 0:
+                    response_file = response_file[0]
+                else:
+                    response_file = response_file[i_module-1]
+            response = cp.load(response_file)
+            RESPONSE_SAMPLING = float(response['time_tick'])
+            RESPONSE_BIN_SIZE = float(response['bin_size'])
+            MIN_STEP_SIZE = min(RESPONSE_BIN_SIZE, RESPONSE_SAMPLING*V_DRIFT)
+            MAX_RADIUS = int(response['response'].shape[0] * RESPONSE_BIN_SIZE // PIXEL_PITCH) # assuming y,z in the response file has the symmetry
```
yszhang95: `MAX_RADIUS` is used together with active pixels, which are projected pixels of a 'line' track. Later, when it is used in `get_neighboring_pixels`, there might be some pixels excluded from the search radius as no diffusion is considered.

diaza:
Hi Yifan!

> Adjusted the MIN_STEP_SIZE which controls the segment chopping the current calculation stage. I previously increased it to roughly match the response bin size, but later realised that it should be finer to represent the diffusion distribution. (0.0064cm -> 0.001 cm) Changed it back to 0.0064cm. Before the seed for diffusion is fixed (the problem and the solution discussed in PR#279), there is no point to reduce is the step size.

I think this is a misunderstanding, which you also mention in this [comment](https://github.com/DUNE/larnd-sim/pull/279#issuecomment-3041007230). For each step, the index passed `itrk * npix * nt + ipix * nt + it ` is not the seed but the index of the rng state to use. After each call of `xoroshiro128p_normal_float32`, the rng state at the index is advanced. 

> I tested with MiniRun5 file123 as an input. It chokes on memory with module1 event 181 by setting up signals with size (n_segments, max_number_of_neighbouring_pixels, time ticks) with or without light simulation.

Could you try merging in `develop` and trying again? I merged in the reproducibility branch to `develop`, which also includes a sub batching where `signals` is calculated.

wforeman:
Hi @YifanC, I haven't fully digested all this work yet but this looks great. A couple small questions that you may be able to answer quickly:

1. With the updated treatment of 't0', is the delay cut-off (now set to 30us) still necessary for memory control? There were recent issues when the neutron source team were trying to run simulations, since nearly all their neutron capture signals will be highly delayed.

2. Have you tested on both beam and non-beam MC? I thought the T0 was calculated differently for them, and thus the cutoff may have an adverse effect for non-beam sims (ie, thinking again of the AmBe neutron simulations).

Thanks!

larndsim/detsim.py
YifanC: `shift_t_collect` is converted from `z` which already included the diffusion. In addition, this if statement only checks the time index `this_time + shift_t_collect` stays in the response bin boundary. Therefore, I think it's correct as it is.

larndsim/detsim.py
YifanC: Yes! Sorry... I casted them to be the same dtype but still very dangerous operation... sorry changing it to check the precision within 1E-6

YifanC:
Hi @wforeman, sorry I was not very clear.
1. Yes, I think one would still need a t0 cut off. Although `signals` no longer depends on t0, `pixels_tracks_signals` (summed signals per pixel) still depends on t0 to use them to place them in the same "readout time frame". Without t0 cutoff, `pixels_tracks_signals` can blow the memory for the same reason. 
2. I'm not sure I understand the differences of t0 between the beam and non-beam cases. They are all assigned at the edepsim stage with respect to the start of events, naively speaking. In the beam case, the event start time got moved, but the treatment for both cases are the same once the beam triggers are subtracted. It will cause problems for neutron simulations for example, as this cutoff will remove some of the decay products. The naive operation here assumes the late signals are small enough to not generate a signal (not validated) but it can cause perturbation to the "main" signals. The cutoff at 30us is a proposal. One can enlarge it in the detector parameter configuration in different runs. However, I think this is brute force. To potentially improve it, if all signals are beam plus a small time delay, or if all signals are delayed, we can factor out the earliest t0 and add it back in later. I think you would need to do multiple passes if the signal span are large (the t0's of the segments are "very different")

larndsim/detsim.py
YifanC: There might be a misunderstanding here.  By definition, `NEIGHBORING_PIX_DIST` is associated with `MAX_RADIUS` (see [lines](https://github.com/DUNE/larnd-sim/blob/75ba73f5879f8f48fa8aeabed06fd69583e8f7ee/larndsim/consts/detector.py#L410-L415)). In addition, `NEIGHBORING_PIX_DIST` is not a single value, but a list of possible distances covered by the range of the response LUT in a unit of pixel pitch. Let me know if I misunderstood your comment.

larndsim/pixels_from_track.py
YifanC: I don't understand.

larndsim/detsim.py
```diff
@@ -141,7 +150,8 @@ def tracks_current_mc(signals, pixels, tracks, response, rng_states):
             # In order to conservatively include more time ticks
             # we use the longest response time, and shortest distance to the cathode from the segments
             # the distance is converted to time using nominal drift velocity
-            if (this_time - t['t0']) > (detector.RESPONSE_MAX_TIME - dist_cathode / detector.V_DRIFT):
+            # pad with 5 times of longitudinal diffusion
+            if this_time > (detector.RESPONSE_MAX_TIME - dist_cathode / detector.V_DRIFT) + t['long_diff'] * detector.DIFF_N_SIGMAS:
```
diaza: Are the units right? Should `t['long_diff'] * detector.DIFF_N_SIGMAS` be divided by detector.V_DRIFT?

larndsim/consts/detector.py
YifanC: I half agree with what you said. Indeed you can have edge case that the main line is 4 pixels away, and with diffusion you can be within the 4 pixels radius (4 is the number used in the current setup and can be replaced.) I think given the diffusion size for the full drift vs. pixel size, this is highly unlikely to happen in large scales. On the other hand, the signal on the 4th neighbour is very small and is unlikely to generate a hit by itself, but can affect a busy environment. If it's really necessary, we can set neighbour radius to be (response coverage + 1 pixel). I don't think it's worth the memory tradeoff in the current simulation logic.

larndsim/detsim.py
```diff
@@ -150,10 +160,13 @@ def tracks_current_mc(signals, pixels, tracks, response, rng_states):
             direction = (segment[0]/length, segment[1]/length, segment[2]/length)
             sigmas = (t["tran_diff"], t["tran_diff"], t["long_diff"])
 
+            # full response range and 5 sigmas of transverse diffusion
             impact_factor = sqrt(response.shape[0]**2 +
-                                     response.shape[1]**2) * detector.RESPONSE_BIN_SIZE
+                                     response.shape[1]**2) * detector.RESPONSE_BIN_SIZE + t['tran_diff'] * detector.DIFF_N_SIGMAS
```
diaza, line 165: Doesn't `sqrt(response.shape[0]**2 +
                                      response.shape[1]**2) * detector.RESPONSE_BIN_SIZE `
already give you the max possible range that can be simulated?

larndsim/consts/detector.py
diaza: Doesn't `response['response'].shape[0]` already give the maximum distance that can be simulated since that's what the response file simulated? Would increasing `MAX_RADIUS` make a difference?

YifanC:
Hi @yszhang95, 

> --- Fix in hit level.
> 
> Bullet 6 deals with signal summation on the pixel pad and backtracking. It walks the track history to identify which contributions belong to each pixel. The key issue is the presence of a spatial window that restricts how far back we track. As long as this window remains, the fix is only partial, and the underlying inaccuracy in hit attribution persists. To fully resolve the issue, the window must be removed entirely. With that in mind, it becomes less motivated to retain the changes in bullet 1 while removing bullet 6.
> 
> On the other hand, if memory usage remains a concern and there’s a possibility we may not keep the edits in bullet 6, it may make sense to reconsider the extent of the changes in bullet 1 as well. In particular, we could explore aligning the neighboring pixel coverage in bullet 1 with the spatial window still present in bullet 6. This would maintain consistency between signal summation and backtracking, and help contain memory growth.
> 
A summary of the relationship between B1 and B6 is that if we don't change B1 (then we are only looking at 1 neighbour. The same argument stands if we increase it to two neighbours), B6 is not needed. We can also keep B1 and remove B6, but than there will be some inefficient calculation that will never be used downstream, which we will still see some improvement. Or we can have B1 and B6 more agreeable in the range with each other, but not sure I am committed to patch that.

> --- Decision in backtracking of truth
> 
> Back-tracking in the existing algorithm can generate misleading “positive” hits, since only the positive portions of a bipolar signal can trigger a hit—even when no charge is actually collected on that pixel. How we choose to record truth information should depend on downstream needs. For example, reconstruction teams may prefer to see only the genuinely collected (“true”) positive hits, rather than these “false” positives.

I agree with what you said in principle we just care about backtracking in the "active pixels" directly under the projection of the segments which correspond to the main pixel in the response. In that sense, our current scheme for backtracking already include these unnecessary neighbours, because the unfortunate entangle of the pixel signal summation and the backtracking range. Therefore, you would have what we have now that some segments can contribute to a hit by >100% and some segments can contribute to a hit by <0%. Ignore this for a second, to me it is still "a philosophical dilemma" that if you have a neighbouring pixel with some induce current. If the positive part of a induced signal generate a hit, would you eventually want to remove this hit and therefore the backtracking? If we agree these false positive hits do represent the real setup and is something we want to remove, then the backtracking should be only on active pixels. If the "induction signal" are mixed with a collective signal, do we track the induction truth contribution? If not, we only need track truth backtracking and renormalise the contribution to 100% (or not, depending on how we want to associate the observed hit information with the truth contribution). However, I feel it makes slightly difference in considering that in for example the truth labeling for SPINE training sample or so, which I would need to double check.

larndsim/detsim.py
YifanC: You are correct! Making that change.

larndsim/detsim.py
YifanC: The `impactor_factor` gives a radius of what you would chop from an original segment to a relevant segment to this particular pixel (it jumps between the segment based and pixel based view...). While the diffusion is only added after you obtain this relevant segment, you can get a part of the original segment that diffuse into the response range, but you have thrown away from this "relevant segment" chop up. That's why you need to include diffusion here.

YifanC:
Hey @diaza 
> Could you try merging in develop and trying again? I merged in the reproducibility branch to develop, which also includes a sub batching where signals is calculated.
Thanks for merging the develop into this branch! The particular instance of the memory choke coming from the creation of `signals` ([here](https://github.com/DUNE/larnd-sim/blob/e244a4e3ac1ad9ee3c93b5cc23f85680e9a4373b/cli/simulate_pixels.py#L1093-L1095)), and the sub-batch is downstream of it. Regardless, I tested it and it has the same problem.

YifanC:
@yszhang95 @diaza 
On a separate note, I tested to keep everything as it is as in this PR but reduced the neighbouring range to 3 (redefining `MAX_RADIUS`) instead of 4 defined by the response LUT. It did run through. Since the reproducibility of the simulation is fixed, maybe someone can test the impact of the neighbouring pixels in the beam setup. The run time without light simulation is 357.48 s.
An alternative way to bypass this issue in the current code setup is to reduce the segment batch size. For the same file, I tested to reduce the batch size from 10000 to 5000, and it ran. It would potentially cause inaccurate pixel signals as some pixels may not have all segment contributions in the same batch, but it's the same issue ND-LAr simulation is facing. It seems that there's only one event in one module that entered the sub-segment batch. The run time without light simulation is 441.61 s
After all, for this scalability issue, if we are keeping the same simulation logic, we need to give up some accuracy. Note that I have only tested with one single file (file 123). There might be other edgy cases occur in the production...

YifanC:
@diaza 

> > Adjusted the MIN_STEP_SIZE which controls the segment chopping the current calculation stage. I previously increased it to roughly match the response bin size, but later realised that it should be finer to represent the diffusion distribution. (0.0064cm -> 0.001 cm) Changed it back to 0.0064cm. Before the seed for diffusion is fixed (the problem and the solution discussed in PR#279), there is no point to reduce is the step size.

> 
> I think this is a misunderstanding, which you also mention in this https://github.com/DUNE/larnd-sim/pull/279#issuecomment-3041007230. For each step, the index passed itrk * npix * nt + ipix * nt + it  is not the seed but the index of the rng state to use. After each call of xoroshiro128p_normal_float32, the rng state at the index is advanced.

Tested it, and indeed it was my misunderstanding. I didn't realise rng state is update internally in the kernel, so if you repeated call `xoroshiro128p_normal_float32` in a kernal with the same index, it will still produce different numbers. Reduced the MIN_STEP_SIZE to 0.001cm.

larndsim/detsim.py
YifanC: Sorry, I misunderstood the initial comment...response.shape[0] is 45, the bin size is 1/10 of a pixel pitch. This [line](https://github.com/DUNE/larnd-sim/blob/b79ee80e4e4b1f456ecb03cd8094b019e4fd2afb/larndsim/consts/detector.py#L410) defines `MAX_RADIUS`

diaza:
> Hey @diaza
> 
> > Could you try merging in develop and trying again? I merged in the reproducibility branch to develop, which also includes a sub batching where signals is calculated.
>
> Thanks for merging the develop into this branch! The particular instance of the memory choke coming from the creation of `signals` ([here](https://github.com/DUNE/larnd-sim/blob/e244a4e3ac1ad9ee3c93b5cc23f85680e9a4373b/cli/simulate_pixels.py#L1093-L1095)), and the sub-batch is downstream of it. Regardless, I tested it and it has the same problem.

Yeah, this came up in a slack thread (which I thought you were in...). I'll add you now. 
Summary: Matt was able to run this branch, but with an unacceptable jump in memory. We discussed past batching schemes, but Matt found an issue with the "digi-batching" scheme we had earlier decided on. He'll revisit his pixel-batching scheme, which seems to work, and we can review.

diaza:
@YifanC 
> Hey @diaza
> 
> > Could you try merging in develop and trying again? I merged in the reproducibility branch to develop, which also includes a sub batching where signals is calculated.
>
> Thanks for merging the develop into this branch! The particular instance of the memory choke coming from the creation of `signals` ([here](https://github.com/DUNE/larnd-sim/blob/e244a4e3ac1ad9ee3c93b5cc23f85680e9a4373b/cli/simulate_pixels.py#L1093-L1095)), and the sub-batch is downstream of it. Regardless, I tested it and it has the same problem.

Yeah, this came up in a slack thread (which I thought you were in...). I'll add you now. 
Summary: Matt was able to run this branch, but with an unacceptable jump in memory. We discussed past batching schemes, but Matt found an issue with the "digi-batching" scheme we had earlier decided on. He'll revisit his pixel-batching scheme, which seems to work, and we can review.

larndsim/detsim.py
yszhang95: Thanks! I misunderstood the calculation of `NEIGHBORING_PIX_DIST`.

larndsim/pixels_from_track.py
yszhang95: This comment is not valid anymore after I realized `NEIGHBORING_PIX_DIST` is calculated from `MAX_RADIUS`.

larndsim/detsim.py
yszhang95: Thanks for clarification. I was wrong in the original comment.

larndsim/consts/detector.py
yszhang95: Hi @diaza, I think this relates to how `MAX_RADIUS` is used. It is used for checking where induced currents are calculated. The proposal of `MAX_RADIUS` is to include all pixels with [-MAX_RADIUS, MAX_RADIUS] within the pixels under the track projection, without considering the diffusion. The `response['response'].shape[0]` gives the maximum distance when the electron is collected in the center pixel. My opinion is `active_pixels` is not necessary to be "center pixel" for all electrons, given the presence of diffusion.

Yes @YifanC , the case that `active_pixels` do not covering all diffused charges might be edge effects, given existing 2x2 setup. I have no conclusion without a dedicated comparisons.

wforeman:
> 2\. The naive operation here assumes the late signals are small enough to not generate a signal (not validated) but it can cause perturbation to the "main" signals

Not true for neutron captures, which will produce gammas that make visible hits. The average neutron capture time is something like ~200us, so a strict 30us cut will exclude a very large fraction of them. Can we extend this delay cut-off at least to the full drift period of 200us?

YifanC:
> > 2. The naive operation here assumes the late signals are small enough to not generate a signal (not validated) but it can cause perturbation to the "main" signals
> 
> Not true for neutron captures, which will produce gammas that make visible hits. The average neutron capture time is something like ~200us, so a strict 30us cut will exclude a very large fraction of them. Can we extend this delay cut-off at least to the full drift period of 200us?

My opinion is it's probably reasonable to do so for a neutron sample production, but would create a lot of overburden for the general purpose sample production. This can be tuned by a configurable parameter in the yaml. 
To address this more thoroughly, I think it would be better to have a second path of signal formation for the late segments, as they don't generate interference for the early signals, and therefore you don't need to force everything to have a huge size. 
Alternatively (probably easier to realise) making signals and the subsequent matrix jagged array can lift the overburden of setting this time cutoff long

diaza:
Hi all
I've merged in `develop`, and will do some benchmarking and looking at validations today. If all looks good, I'll merge sometime tomorrow unless there are apprehensions.

cli/simulate_pixels.py
```diff
@@ -1219,7 +1215,10 @@ def save_results(event_times, results, i_trig, i_mod=-1, light_only=False):
                 BPG_Y = max(ceil(signals.shape[1] / TPB[1]),1)
                 BPG_Z = max(ceil(signals.shape[2] / TPB[2]),1)
                 BPG = (BPG_X, BPG_Y, BPG_Z)
-                pixels_signals = cp.zeros((len(unique_pix), signals_ticks))
+                # Here inflate the signal_ticks by the track t0
+                # All the late segments have been removed in the loading stage
+                signals_ticks_t0 = signals_ticks + ceil(selected_tracks['t0'].max() / detector.TIME_SAMPLING)
```
diaza, line 1220: @YifanC 
Would it make more sense to have something like 
`signals_ticks_t0 = selected_tracks['t_end'].max() + selected_tracks['long_diff'].max() / detector.V_DRIFT * detector.DIFF_N_SIGMAS`
instead? 
If I'm understanding the code right, this would give the minimum number of ticks needed to encompass the necessary time windows. 
Although maybe the effect would be too minor to consider?

YifanC: In principle, `signals_ticks_t0 = ceil((selected_tracks['t_end'].max() + selected_tracks['long_diff'].max() / detector.V_DRIFT * detector.DIFF_N_SIGMAS)/ detector.TIME_SAMPLING)` works and should give a smaller number, except when you need to consider the misalignment between `detector.RESPONSE_MAX_TIME` and `detector.DRIFT_MAX_TIME`. Perhaps it's better to define the non negative differences `detector.RESPONSE_MAX_TIME - detector.DRIFT_MAX_TIME` and make the expression cleaner.

diaza:
Hey all, here's some benchmarking and validation plots before the merge. 
I merged in PR https://github.com/DUNE/larnd-sim/pull/286 here, since that fix made the validations a little easier. It removed the early hits I was seeing, so I believe that issue is solved. 

For benchmarking, I ran the `develop` branch, and this branch with MAX_RADIUS set to 0,1,2,3, and 4. For each, I ran with PIXEL_BATCH_SIZE set to 2000 (current default), 1000, and 100. I paste a table with the peak memory and run time for each. You can see general trends, run time increases with MAX_RADIUS, and decreases with PIXEL_BATCH_SIZE. Peak memory somewhat increases with MAX_RADIUS, but I'm not seeing the expected decrease as a function of decreasing pixel_batch_size. I'm not entirely sure why this is the case (maybe @mjkramer can comment?), but it could be the means in which I'm measure memory usage (described as "hacky" by Matt). Regardless, the batching has definitely helped in decreasing overall memory usage, and I'm fine merging despite this peculiarity. 
image

I'm attaching validation plots from an output from develop, and another from MAX_RADIUS=4. 
 [develop_default.flow.0000000.FLOW.hdf5_validations.pdf](https://github.com/user-attachments/files/21781939/develop_default.flow.0000000.FLOW.hdf5_validations.pdf)
[feature_batch_default_max_radius_4.flow.0000000.FLOW.hdf5_validations.pdf](https://github.com/user-attachments/files/21781940/feature_batch_default_max_radius_4.flow.0000000.FLOW.hdf5_validations.pdf)
The main differences of note are that the reco vs true positions are more spread out (which makes sense to me), and a bigger spread in the sum of charge fraction for each hit (makes less sense to me).

Unless there are concerns, I'm fine merging this in by the end of Friday.

PR #284, opened 2025-07-25
Pixel batching

mjkramer:


diaza:
Hey @mjkramer, thanks for all of this! 
Let me make sure I understand what happens before and after.

Before this PR:
1. We start with a list of segments. 
2. This list of segments can be batched if it is very long. 
3. For each segment within each batch, we get all the "relevant" pixels. 
4. We then simultaneously find the induced current for each element in the (segment x pixel x time) matrix
5. Combine currents and digitize
The above is unideal since batching the segments will have the batches run through to the digitization step independently. 

My understanding of your PR:
1. We start with a list of segments.
2. For each segment, you get all the "relevant" pixels. 
2. You then make an inverse map, where you get 2d array where the first dimension are the unique pixels, and the second element are the inducing segments for each pixel.
4. You batch over these pixels.
5.  For each batch, you get the relevant segments 
6.  We then pick up from step 3 in the previous scheme 

Do I understand those two schemes right? 
In the new batching scheme, there's the step where we get the pixel batch, then we get the segments associated with that pixel batch, and then get the pixels that those segments can induce a signal on. This second set of pixels would be a superset of the pixels in the batch. Do we make sure to record only the pixels selected for the batch? I think so because of this [line](https://github.com/DUNE/larnd-sim/blob/a787f52b953bb5ef982453a0144344b33a150b14/cli/simulate_pixels.py#L1120), but I want to double check?

mjkramer:
Hi @diaza, thanks for taking a look. Your understanding agrees with mine, and I have the same question as you:

> In the new batching scheme, there's the step where we get the pixel batch, then we get the segments associated with that pixel batch, and then get the pixels that those segments can induce a signal on. This second set of pixels would be a superset of the pixels in the batch. Do we make sure to record only the pixels selected for the batch? I think so because of this [line](https://github.com/DUNE/larnd-sim/blob/a787f52b953bb5ef982453a0144344b33a150b14/cli/simulate_pixels.py#L1120), but I want to double check?

This (stated better than I've been able to) is exactly the reason I've been hesitant to merge these changes. Our selected segments could induce signals on pixels outside of the pixel-batch (since those "extra" pixels could end up in `neighboring_pixels`), and there's nothing in the call to `tracks_current_mc` that would prevent this. The hope is that the downstream functions will discard the signals for those pixels (since those pixels are absent from `unique_pix`, which does get passed to some of these functions), so that the only end-effect is that we've wasted a bit of computation on pixels that will get "officially" processed in a different batch.  However, this hope needs to be confirmed.

Alternatively, we could think about modifying `tracks_current_mc` to make use of `unique_pix`. Or we could do a cleanup pass on  `neighboring_pixels` and `active_pixels` to remove any that are absent from `unique_pix`. Yeah... if we clean up the pixel arrays like that, then we won't be reliant on assumptions about the implicit behavior of downstream functions, so I'm leaning in that direction...

diaza:
Reading through the code, I feel fairly confident that everything sorts itself out. `pixel_index_map` is constructed using only `unique_pix`. In `get_track_pixel_map2`, only pixel ids within `unique_pix` are [considered](https://github.com/DUNE/larnd-sim/blob/e56a1281e46b88f10a5d0ed48aa3e8bd9669bf45/larndsim/detsim.py#L328) when constructing `track_pixel_map`.  Finally `sum_pixel_signals` uses `pixel_index_map` and `track_pixel_map` when making `pixels_signals`.

But I agree that it'd be more comforting if we just clean up the pixel arrays. Maybe right [after ](https://github.com/DUNE/larnd-sim/blob/6712274be2bd90a63fa70a79887216d660bcb645/cli/simulate_pixels.py#L1120)`get_pixels`, we filter out where `neighboring_pixels` and `active_pixels` has elements not in `unique_pix`. We would then also zero out the corresponding values in `neighboring_radius`. I'm not sure if `n_pixels_list` is ever used? I also don't think we ever use `active_pixels`? 

I'm also wondering if the end result code could be made simpler if we rearrange `tracks_current_mc` and subsequent functions so that `signals` has pixels in the first dimension instead of the second. That way we wouldn't have to do all this flip flopping between segments<->pixel maps, and it'd be easier to batch things since the first dimension is pixels (since each pixel is entirely independent of the others). Thoughts on that?

diaza:
Also, are `invert_array_map` and `get_track_pixel_map2` redundant? I'll have to double check when I have the time.

diaza:
@mjkramer, I've added the bit of code that gets rid of superfluous pixels. If things look good, feel free to merge. 

I'm attaching two sets of validation plots. The first is from a run with `develop`, and the second is from a run from this branch, with the pixel batching set to the small value of 100. Plots look good, and I got a memory reduction of 40%, with a 20% increase in run time. 
[develop.flow.0000000.FLOW.hdf5_validations.pdf](https://github.com/user-attachments/files/21674872/develop.flow.0000000.FLOW.hdf5_validations.pdf)
[feature_alex_100.flow.0000000.FLOW.hdf5_validations.pdf](https://github.com/user-attachments/files/21674877/feature_alex_100.flow.0000000.FLOW.hdf5_validations.pdf)


I also ran a test to see if the code already filtered out superfluous pixels. In this test, instead of filtering out the superfluous pixels [here](https://github.com/DUNE/larnd-sim/blob/ab0a35c0fbfedd454c6c93022b755c096048bef6/cli/simulate_pixels.py#L1139), I filter out the pixels in `signals` after the signals are calculated. When I compare the outputs with and without this filter, I find them to be identical, meaning that the set up without the filters was working properly. (I can't do the test with leaving the filter at [L1139](https://github.com/DUNE/larnd-sim/blob/ab0a35c0fbfedd454c6c93022b755c096048bef6/cli/simulate_pixels.py#L1139), since that will change how the RNGs are sampled in `tracks_current_mc`)

Regardless, it's good to have this filter explicitly in the code.

diaza:
Actually, the summed charge fractions plot looks a little weird. Let me check it out

diaza:
When I zoom into the summed fractions plot for the pixel batch of size 100, it looks fine. I only see 3 hits that go above 10. I think it's still fine to merge. (But I will probably spend tomorrow trying to figure out why this happens)
image

mjkramer:
Awesome, thanks for the cross-checks! I'm now much more confident about merging this. Regarding the handful of new "high-sum" hits that popped up in the test you ran, my guess is that the difference arises from the different sampling of the RNGs. Those hits are nothing new and presumably come from cases where:

1. Some segments provide "negative" contributions to the hit due to the bipolar induction signal
2. Then, to ensure that the sum of contributions = 1, the "positive" contributors get scaled up
3. Then we sort the segments by their contributions and truncate the list, throwing away the smallest / most negative contributors, so that the sum of the remaining contributions is now over unity

It may be worth addressing these cases, maybe by using a jagged array to avoid the truncation, or maybe by "renormalizing" the contributions, but I'll leave that for another discussion...

PR #286, opened 2025-08-09
Pre-trigger hits

jaafar-chakrani:
Pre-trigger hits seem to be due to a PACMAN delay adjustment in: https://github.com/DUNE/larnd-sim/blob/3849e3f406734784c15f8860716a2e369a859542/larndsim/fee.py#L680-L681

The hit timestamp (in `adc_ticks_list`) should not be impacted by the transfer delay from the LArPix chip to the PACMAN, but rather the packet `receipt_timestamp`. Here I removed this delay from the `timestamp` to the `receipt_timestamp`.

A more correct implementation would calculate the delay in `receipt_timestamp` as a function of the depth of the chip in the hydra network, but I don't think this information is used in any downstream analyses on the simulation.

### DIFF
--- a/larndsim/fee.py
+++ b/larndsim/fee.py
@@ -269,7 +269,7 @@ def export_to_hdf5(event_id_list,
                             continue
                 p.chip_key = chip_key
                 p.channel_id = channel
-                p.receipt_timestamp = time_tick
+                p.receipt_timestamp = time_tick + 20 * detector.CLOCK_CYCLE
                 p.packet_type = 0
                 p.first_packet = 1
                 p.assign_parity()
@@ -677,8 +677,7 @@ def get_adc_values(pixels_signals,
                 crossing_time_tick = min((ic, len(time_ticks)-1))
                 # handle case when tick extends past end of current array
                 post_adc_ticks = max((ic - crossing_time_tick, 0))
-                #+2-tick delay from when the PACMAN receives the trigger and when it registers it.
-                adc_ticks_list[ip][iadc] = time_ticks[crossing_time_tick]+time_padding-2+post_adc_ticks
+                adc_ticks_list[ip][iadc] = time_ticks[crossing_time_tick]+time_padding+post_adc_ticks
 
                 ic += round(detector.RESET_CYCLES * detector.CLOCK_CYCLE / detector.TIME_SAMPLING)
                 last_reset = ic
--- a/larndsim/fee.py
+++ b/larndsim/fee.py
@@ -269,7 +269,7 @@ def export_to_hdf5(event_id_list,
                             continue
                 p.chip_key = chip_key
                 p.channel_id = channel
-                p.receipt_timestamp = time_tick + 20 * detector.CLOCK_CYCLE
+                p.receipt_timestamp = time_tick
                 p.packet_type = 0
                 p.first_packet = 1
                 p.assign_parity()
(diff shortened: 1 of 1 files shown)


krwood (commented):
I think the original comment is actually misleading what the `-2` (which should really be a `+2`) is trying to do. I don't think this is actually about simulating the time it takes to get the packet off of the tile and to the pacman receivers. Rather, I think this is meant to simulate the fact that it takes 2 clock cycles for the larpix channel to register that the signal is above threshold once it does cross the threshold. We should confirm with the ASIC engineers, but that would make much more sense to me here. If that is the case, then I think we would want to add a `+2` into L680.

diaza:
Hello! @krwood , @russellphysics, and I chatted earlier and I'd like to summarize some relevant points, plus some thoughts of mine. 

First, like Kevin says above, the `-2` does not refer to some delay in the receipt_timestamp, but the time it takes in clock cycles for the channel to register that the charge is above threshold. So we should remove the correction made here 
```python 
p.receipt_timestamp = time_tick + 20 * detector.CLOCK_CYCLE
```
This delay does affect the timestamp though. 

Brooke clarified for us that the timestamp of the pixel is not when the pixel triggered, but instead when the pixel latches, meaning at the end of the integration time. Let me summarize the following lines of [code](https://github.com/DUNE/larnd-sim/blob/3849e3f406734784c15f8860716a2e369a859542/larndsim/fee.py#L681), which assign the time of a packet after an integration window
```python
                crossing_time_tick = min((ic, len(time_ticks)-1))
                # handle case when tick extends past end of current array
                post_adc_ticks = max((ic - crossing_time_tick, 0))
                #+2-tick delay from when the PACMAN receives the trigger and when it registers it.
                adc_ticks_list[ip][iadc] = time_ticks[crossing_time_tick]+time_padding-2+post_adc_ticks
```
`ic` is the index along the time dimension of `pixels_signals` within the `while` loop. (For the following, let's assume the case where` ic < len(time_ticks)-1`, which will give `crossing_time_tick = ic` and ` post_adc_ticks = 0`) . When we are at this point of the code, `ic` is at the end of the integration window, so `time_ticks[crossing_time_tick` will give the time at the end of the integration window. 
Now, if the integration window takes into account the 2 clock cycle delay, then we don't actually have to add any term to `adc_ticks_list`, since it'll be included in `time_ticks[crossing_time_tick]`.
This opened a question between Kevin and I, why are we getting negative times if the integration time is 18 ticks but we are subtracting 2 ticks? The reason is that `time_ticks ` is actually constructed [(here)](https://github.com/DUNE/larnd-sim/blob/3849e3f406734784c15f8860716a2e369a859542/cli/simulate_pixels.py#L1256) in units of us.  
 ```python
                 time_ticks = cp.arange(0, len(unique_eventIDs) * max_signal_time, detector.TIME_SAMPLING)
```
where detector.TIME_SAMPLING  = 0.1 ms.
In other words, when we subtract "2", we're actually subtracting by 2 us, not 2 ticks. I think this explains why we're seeing hits come in 1 tick early. Reading through `get_adc_values`, I've convinced my self that the earliest possible timestamps we can get is at 19 ticks (1.9 ms), from which we subtract 2 us (20 ticks) and end up with hits one tick early. 

In the above snippet, it also handles the case where the time dimension of pixels_signals is longer than `time_ticks`. This shouldn't happen since both arrays' lengths are determined by `max_signal_time`. But, in the event where this does happen, it tries to extrapolate where the ending timestamp should be. It does this incorrectly though, since adc_ticks_list is in units of us, instead of ticks.  

The next open question is how long the integration window should actually be [here](https://github.com/DUNE/larnd-sim/blob/c37b2188d2656de6c126dac4c5bcbce2bdd653ae/larndsim/fee.py#L626)
```python
                interval = round((3 * detector.CLOCK_CYCLE + detector.ADC_HOLD_DELAY * detector.CLOCK_CYCLE) / detector.TIME_SAMPLING)
```
There's the integration window `detector.ADC_HOLD_DELAY`, and then an extra `3 * detector.CLOCK_CYCLE`. Does this extra term take into account the delay in the trigger? If so, then we don't have to add any time to `adc_ticks_list `. Kevin also mentioned that there can be a tick's worth of charge integrated after the pixel latches (did I get that right?); is that included in  3 * detector.CLOCK_CYCLE? If so, should that also contribute to the `timestamp`, or should it be subtracted from `adc_ticks_list`? Brooke is looking into this.

diaza:
If people agree with my assessment of not adding anything to `adc_ticks_list`, I think we can merge this in for MR6.5 and leave the question of integration time for later?

diaza:
@jaafar-chakrani @krwood 
FYI, I changed the target branch from `develop` to `small_fixes`, and merged.

PR #287, opened 2025-08-10
Updated 2x2 thresholds

jaafar-chakrani:
2x2 thresholds are updated to take into account pre-beam run tunings. 

These are the thresholds used in MR6.4: 
old_thresholds

And below are the updated thresholds:
new_thresholds


krwood:
Just a small clarification for the record: At the time of writing we have not run the larnd-sim stage of MR6.5. The "old threshold" Jaafar shows in his first plot were the ones used in *MR6.4*.

krwood:
@jaafar-chakrani there was also a report from @seg188 that there seems to be a tile swap somewhere in the threshold application. I think it was tiles 7 and 8 on io_group 6. @seg188 can you comment on your observations and whether you think there needs to be something changed in larnd-sim or elsewhere?

diaza:
@krwood 
> @jaafar-chakrani there was also a report from @seg188 that there seems to be a tile swap somewhere in the threshold application. I think it was tiles 7 and 8 on io_group 6. @seg188 can you comment on your observations and whether you think there needs to be something changed in larnd-sim or elsewhere?

I've resolved this issue in this PR https://github.com/DUNE/ndlar_flow/pull/205

PR #289, opened 2025-08-28
Hotfix backtracking fix

diaza:
This PR corrects the backtracking issues observed by Bruce Howard  in the initial MR6.5 spin. He observed that there were many hits without any backtracked segments. Here is the Slack [link](https://dunescience.slack.com/archives/C0496EJK59D/p1756154331847469) to his post, and I've attached his slide. 
[2x2_MiniRun_Comparison.pdf](https://github.com/user-attachments/files/22033838/2x2_MiniRun_Comparison.pdf)

The issue was traced first to bad `current_fractions` (Slack link [here](https://dunescience.slack.com/archives/C0496EJK59D/p1756272257800459)) and ultimately to `pixels_tracks_signals` ([here](https://github.com/DUNE/larnd-sim/blob/1ba10107de3192f020de5ec36e8c14a3668c25c9/cli/simulate_pixels.py#L1231)) being filled in incorrectly. 

This was tested by comparing `pixels_tracks_signals` and `pixels_signals`. `pixels_signals` (shape (#pixels, #ticks))gives the combined induced signal on each pixel. `pixels_tracks_signals` (a 1D array which can be conceptualized as a jagged array of shape (#ticks, #pixels, #segments), where #segments is the jagged component) gives the induced signal on each pixel from each segment. Therefore `pixels_tracks_signals` can be compared with `pixels_signals` to check if it is being filled in properly. 

[After](https://github.com/DUNE/larnd-sim/blob/1ba10107de3192f020de5ec36e8c14a3668c25c9/cli/simulate_pixels.py#L1239) `sum_pixel_signals`, I collapse `pixels_signals` and `pixels_tracks_signals` so that it gives me the total induced signal on each pixel and print that out (code pasted below for reference)
```python
                 print("pixel_signals\n", pixels_signals.sum(axis=-1))                                               
                                                                                                                     
                 a = pixels_tracks_signals.get()                                                                     
                 a = np.reshape(a, (signals_ticks_t0, int(num_backtrack.sum())))                                     
                 bt  = cp.append(offset_backtrack, [offset_backtrack[-1] + num_backtrack[-1]])                       
                 bt = bt.get()                                                                                       
                 a = [ a[:, bt[i]:bt[i+1]].sum(axis=-1) for i in range(len(bt)-1)]                                   
                 a = np.array(a)                                                                                     
                 print("\n pixels_tracks_signals\n", a.sum(axis=-1))
```
An example output is pasted below, showing inconsistencies. 
```
pixel_signals:
 [ 0.94027581 -3.17951654  0.72214372 ...  0.          0.
  0.        ]

 pixels_tracks_signals:
 [ -1.01580912  -9.14673328 -26.20556638 ...   0.40930901  -0.27295125
   0.34227214]
```

The issue comes from [this](https://github.com/DUNE/larnd-sim/blob/1ba10107de3192f020de5ec36e8c14a3668c25c9/cli/simulate_pixels.py#L1241) line, where the `t0` added back in is being passed in as 
`cp.array(selected_tracks['t0'])`
when it should be `cp.array(selected_tracks['t0']/detector.TIME_SAMPLING, dtype = int)`. When `pixels_tracks_signals` is being filled, the incorrect `track_t0` [here](https://github.com/DUNE/larnd-sim/blob/1ba10107de3192f020de5ec36e8c14a3668c25c9/larndsim/detsim.py#L258) will put the starting index in the wrong location. The wrong scale will shift the start of the signal down by a factor of 10, and non-integer `track_t0` will place `base_idx` in the middle of the conceptual (pixel, segment) sub-array, instead of the beginning. (I can go in more detail if someone would like).

 With the commits above, the output is now 
 ```
pixel_signals
 [ 0.94027581 -3.17951654  0.72214372 ...  0.          0.
  0.        ]

 pixels_tracks_signals
 [ 0.94027581 -3.17951654  0.72214372 ...  0.          0.
  0.        ]
```

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -1191,6 +1191,9 @@ def run_simulation(input_filename,
                 
                 # Now directly map neighboring_pixels to pixel indices using lookup
                 pixel_index_map = pix_lookup[neighboring_pixels]
+                # Some elements of neighboring_pixels can have values of -1.
+                # We want to make sure these pixels are also removed in pixel_index_map.`
+                pixel_index_map[neighboring_pixels==-1] = -1
                 RangePop()
 
                 RangePush("track_pixel_map")
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -1238,7 +1238,7 @@ def run_simulation(input_filename,
 
                 detsim.sum_pixel_signals[BPG,TPB](pixels_signals,
                                                   signals,
-                                                  cp.array(selected_tracks['t0']),
+                                                  cp.array(selected_tracks['t0']/detector.TIME_SAMPLING, dtype = int),
                                                   pixel_index_map,
                                                   track_pixel_map,
                                                   pixels_tracks_signals,
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -1223,7 +1223,7 @@ def run_simulation(input_filename,
                 # num_backtrack[ipix] is the number of segments contributing to the pixel
                 num_backtrack = cp.sum(track_pixel_map != -1, axis=-1)
                 # pixels_tracks_signals is a jagged array of conceptual dimension
-                # (#unique_pix, #ticks, backtracked_segments)
+                # (#ticks, #unique_pix, backtracked_segments)
                 # where the final axis (over segments) is jagged.
                 # Physically it's represented as a 1D array where the time index
                 # increments the slowest, followed by the pixel index, followed
--- a/larndsim/detsim.py
+++ b/larndsim/detsim.py
@@ -125,6 +125,7 @@ def tracks_current_mc(signals, pixels, tracks, response, rng_states):
     if itrk < signals.shape[0] and ipix < signals.shape[1] and it < signals.shape[2]:
         t = tracks[itrk]
         pID = pixels[itrk][ipix]
+        if pID < 0 : return
         pID_x, pID_y, pID_plane = id2pixel(pID)
 
         if pID_x >= 0 and pID_y >= 0:


krwood:
Very interesting and a great catch Alex! This seems like a pretty important bug that has somehow snuck by until now. I wonder if this is also responsible for some of the strange behavior we've been seeing for signals near the anode (cc @russellphysics). Could potentially also have an impact on @Rossd437 's electron lifetime extraction?

diaza:
Hey @krwood. I actually don't think this should have affected things pre-MR6.5. @YifanC can correct me, but this "t0 padding" (explained below) was introduced in the recent PR `small_fixes` ("small" is an understatement). 

Regarding t0 padding, this was a memory saving technique by Yifan to not bother creating time ticks in the `signals` array (a very chunky array) for the time between the trigger and the deposition of the first segment. These ticks have to be added in later, which is where the bug was introduced. 

 I've got another bug I found I want to fix. After that I'll make the validation plots and tag some people.

YifanC:
Oops, sorry... Thank you @diaza! Definitely a easter bug I planted... I didn't test it extensively, sorry (Perhaps good reasons to have CI tests..)
@krwood The error should only affect MR6.5 and nothing before.

diaza:
I pushed two more commits. 
I add [this](https://github.com/DUNE/larnd-sim/blob/d57088bd64f190f8503bc398cfe94501035517e9/larndsim/detsim.py#L128) filter in `tracks_current_mc`, since it would still run if pID = -1 and fill (I think) in the `pID_x = detector.N_PIXELS[0] - 1` and `pID_y = detector.N_PIXELS[1] - 1`. Without this fix (and the one [here](https://github.com/DUNE/larnd-sim/blob/d57088bd64f190f8503bc398cfe94501035517e9/cli/simulate_pixels.py#L1196)), I was getting a few extra hits that was messing with my validations. 

I expect the pID filter might slightly speed up the code. 

I'm attaching validation plots. Of interest are the fraction sum plots, and the number of backtracked segments per hit plots. 
 
[develop.0000000.FLOW.hdf5_validations.pdf](https://github.com/user-attachments/files/22069165/develop.0000000.FLOW.hdf5_validations.pdf)
[feature3.0000000.FLOW.hdf5_validations.pdf](https://github.com/user-attachments/files/22069166/feature3.0000000.FLOW.hdf5_validations.pdf)

YifanC:
Thank you @diaza. It makes sense to me! I just have some questions about the plots. What does the plot mean by "sum of packet fractions for each calib hit"? Why there are so many have 0 and >1 sum?

diaza:
@YifanC 
The plot gives the sum of the `fractions` field for each calib prompt hit. You're right it should ideally add up to ~1 for each hit, but it's been a long standing issue (see plots from MR6.2 [here](https://portal.nersc.gov/project/dune/data/2x2/simulation/productions/MiniRun6.2_1E19_RHC/MiniRun6.2_1E19_RHC.plots/PLOTS/FLOW/0000000/MiniRun6.2_1E19_RHC.flow.0000000.FLOW.hdf5_validations.pdf)). 
Deviations from exactly 1 are expected, for example negative contributions will probably be truncated out, giving >1 for the sum. Also, when the fractions are normalized [here](https://github.com/DUNE/larnd-sim/blob/be0b6169263e53ee17363c69399625921a119562/larndsim/fee.py#L673), the numerator is the "measured" charge (including noise and other statistical fluctuations), while the denominator is the "true" charge. This could also lead to deviations from 1. 
But, overall, I don't know if the above factors can explain the scale of the observed deviations from 1.

YifanC:
@diaza Thanks. Yea, the contribution from the negative current should be ~1% at most. I looked at a few pixels and the fractions are weird. If `get_adc_values` is doing what it is suppose to do, then the sum of different segments in `pixels_tracks_signals` and `pixels_signals` have a mismatch for these sum frac significantly deviate from 1. To get 0 contribution, you also need to have `pixels_tracks_signals` all have 0 contributions from any segments for that pixel, which shouldn't ever happen I think?

diaza:
@YifanC 
I agree it's weird. It's not obvious to me where these are coming from and should be looked into. I think, though, that this is a separate issue from the PR, so I'll merge and we can follow up on the weird fractions elsewhere.

PR #290, opened 2025-08-29
Restoring delayed segment t0 filter

wforeman:
While investigating catastrophic memory failures in MiniRun6.5 jobs, where CUDA is asked to allocate O(TBs) resulting in the job crashing, I discovered that this was due to segments with highly-delayed `t0` no longer getting removed like they were in tag v0.4.2. It looks like this code was re-arranged slightly since that tag. This PR moves the delayed t0 filter back to where it was before, filtering out the segments as early as possible in the code before any subsets or copies of the `tracks` array can be created. With the change in place, these memory crashes no longer occur.

I found that the 'detector' properties parameter `SIGNAL_OVERLAP_CUT` couldn't be accessed in this part of the code, so I made a new parameter `MAX_SEGMENT_T0` to `consts/sim.py` to control this filter instead. It is set to the same value of 30 microseconds.

It also adds a filter removing segments from neutral particles as well (neutrons and gammas) since they will not produce ionization signals. This reduces the peak memory slightly by 5-10%.

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -306,8 +306,6 @@ def run_simulation(input_filename,
 
     print(LOGO)
     print("**************************\nLOADING SETTINGS AND INPUT\n**************************")
-    
-    
 
     if not os.path.exists(input_filename):
         raise Exception(f'Input file {input_filename} does not exist.')
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -306,6 +306,8 @@ def run_simulation(input_filename,
 
     print(LOGO)
     print("**************************\nLOADING SETTINGS AND INPUT\n**************************")
+    
+    
 
     if not os.path.exists(input_filename):
         raise Exception(f'Input file {input_filename} does not exist.')
@@ -591,6 +593,20 @@ def run_simulation(input_filename,
             tracks['t0_end'] = tracks['t0_end'] - localSpillIDs*sim.SPILL_PERIOD
             tracks['t0'] = tracks['t0'] - localSpillIDs*sim.SPILL_PERIOD
 
+        # Filter out neutrons and gammas, which will not directly create visible charge or light
+        # (excluding these segments here results in a modest ~10% improvement to memory usage later on,
+        # since this reduces the size of the arrays CUDA must inialize for pixel current calculations)
+        neutrals_mask = (tracks['pdg_id'] != 2112) & (tracks['pdg_id'] != 22)
+        if sum(~neutrals_mask) > 0: print("Rejected ",sum(~neutrals_mask), "track segments from neutral particles")
+        tracks = tracks[neutrals_mask]
+
+        # Filter out highly-delayed segments
+        t0_delay_mask = (tracks['t0'] < sim.MAX_SEGMENT_T0)
+        if sum(~t0_delay_mask) > 0:
+          print("Rejected ",sum(~t0_delay_mask)," highly-delayed segments with T0 > ",sim.MAX_SEGMENT_T0," us: ")
+          for val in tracks[~t0_delay_mask]: print(' t0 = ',val['t0'])
+        tracks = tracks[t0_delay_mask] 
+
         if 'segment_id' in tracks.dtype.names:
             segment_ids = tracks['segment_id']
             trajectory_ids = tracks['file_traj_id']
@@ -757,15 +773,7 @@ def run_simulation(input_filename,
     active_tracks = all_mod_tracks[active_tracks_mask]
     active_segment_ids = all_mod_segment_ids[active_tracks_mask]
     active_trajectory_ids = all_mod_trajectory_ids[active_tracks_mask]
-
-    t0_delay_mask = (active_tracks['t0'] < detector.SIGNAL_OVERLAP_CUT)
-    tracks = active_tracks[t0_delay_mask]
-    segment_ids = active_segment_ids[t0_delay_mask]
-    trajectory_ids = active_trajectory_ids[t0_delay_mask]
-    end_mask = time()
-    print(f"{len(all_mod_tracks) - len(active_tracks_mask)} segments are removed due to the active volume cut. \
-            In addition, {np.count_nonzero(~t0_delay_mask)} segments are removed due to the t0 delay cut.")
-    print(f" {end_mask-start_mask:.2f} s")
+    print(f"{len(all_mod_tracks) - len(active_tracks_mask)} segments are removed due to the active volume cut.")
 
     # We need to make cupy arrays of these and pass them to the kernels;
     # otherwise numba will try to use the GPU's "global constant" memory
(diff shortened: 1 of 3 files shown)


diaza:
I haven't had the time to look at this request yet, but I think @YifanC had removed the lines you're adding back in. Yifan, could you comment? Was that removal intentional or some merging error?

wforeman:
Yes, @YifanC had just moved the lines to slightly later in the code, and modified the approach -- ie, filtering a batched subset of the tracks... though at that point the track collection is already splintered into several different copies like `active_tracks`, `all_mod_tracks`, etc. I think the intention was to preserve the same functionality, but things get a little complicated at that point and I couldn't figure out why these delayed segments were still making their way into the final current calculation arrays. So I just reverted to the original filter which I think makes things simpler. @YifanC - was there a rationale for moving the filter to the location you did?

YifanC:
Hi @wforeman, thank you for taking care of it. 
I modified such because it was initially implemented under the condition of beam simulation, so I just moved out to allow the cut been applied to all kinds of the simulation. I shifted it next to the fiducial volume cut intending to keep all segment cuts in one place. Just to clarify, it was before looping of modules and potential batching of segments/pixels.
In general, I think this PR is good to go. If we will merge this, I would suggest to clean up `SIGNAL_OVERLAP_CUT` in consts/detector.py.
However, I actually don't understand the initial message, which refers to it does not remove delayed segments and `SIGNAL_OVERLAP_CUT` cannot be accessed there. @wforeman Do you mind elaborating? I ran a simple test, it seems `detector.SIGNAL_OVERLAP_CUT` is defined there? Does the print out for these files with delayed segments saying there are no segments removed? Do you mind sharing a file path that has late segments?
I have an additional question. I suppose MR6.5 uses the same edep files as previous MR's. Are the memory explosion occurring on the same files? Were they try to allocate O(100GB) memory?

wforeman:
@YifanC - thanks for the reply! Yes, in this new PR I similarly moved it out of the `IS_SPILL_SIM` loop so it will apply to all sim types, as your fix did.

MR6.5 uses the same edepsim outputs as prior MiniRuns so I am inclined to believe these are the same files that were causing issues before as well when we were spinning MR6.4.

Here is a file I've been testing on, with some delayed segments (2 with T0 ~ 1476133592, and another 27 with T0 ~ 36):
`/pscratch/sd/d/dunepr/output/MiniRun6.5/run-convert2h5/MiniRun6.5_1E19_RHC.convert2h5/EDEPSIM_H5/0000000/MiniRun6.5_1E19_RHC.convert2h5.0000941.EDEPSIM.hdf5
`

Regarding `detector.SIGNAL_OVERLAP_CUT` variable not being accessible, here's an example. If I try to add this at [line 604](https://github.com/DUNE/larnd-sim/pull/290/commits/69943f856512dc3d3acc06aba3a6bf48089200c6#diff-e9e04e7ed8e6419135d41be6955588c04230874bf4fa7e874c445c75a00d7246R604):
<pre>
 t0_delay_mask = (tracks['t0'] < detector.SIGNAL_OVERLAP_CUT)
 </pre>
 
 I get this error: 
 
 `  File "/pscratch/sd/d/dunepr/software/MiniRun6.5/ND_Production/run-larnd-sim/larnd.venv/bin/simulate_pixels.py", line 603, in run_simulation
    t0_delay_mask = (tracks['t0'] < detector.SIGNAL_OVERLAP_CUT)
                                    ^^^^^^^^
UnboundLocalError: cannot access local variable 'detector' where it is not associated with a value`

Do you (or others) know why this might be happening for me? Personally I think this variable might make more sense under the "sim" category anyway, since it isn't inherently tied to any detector readout parameters and the filtering happens purely at truth level before any detector effects are applied. So I'm inclined to keep it in consts/sim.py (but still curious why it doesn't work for me in detector.py).

YifanC:
Hi @wforeman, I think your PR is good after cleaning up the unused the `detector.SIGNAL_OVERLAP_CUT`. 
I don't insist to keep `detector.SIGNAL_OVERLAP_CUT`, but I still don't understand the problem. 
If you add it line 604, it wouldn't work, because `larndsim.consts.detector` hasn't been imported while `larndsim.consts.sim` has. It makes sense. 
However, using the file you pointed, I ran it with the develop branch which I suppose the MR6.5 is based on. I see the following messages which indicate that the late segments are removed. 
```Skipping non-active volumes...15 segments are removed due to the active volume cut.             In addition, 37 segments are removed due to the t0 delay cut.```
I ran larndsim with the input without light simulation on a 40GB A100, and did encounter the memory issue on event 50 in module0, while allocating `signals` [here](https://github.com/DUNE/larnd-sim/blob/be0b6169263e53ee17363c69399625921a119562/cli/simulate_pixels.py#L1152-L1154). What I see, is that the number of segments exploded for this pixel, which I don't know if it's feasible. It only has 414 neighbouring pixels, but 10643 selected tracks. I tried to reduce the pixel batch to 500 from 2000, but I think given the number of neighbouring pixels, it probably didn't hit limit. Tagging @mjkramer here to see if there's a suggestion to deal with rather than further reduce the pixel batch. In total this event has 55785 segments. Removing neutron and photons, it still has 47663 segments. Further removing nuclei, it still has 47202 by these pdg (-321,       -211,        -13,        -11,         11,        211,       2212). I don't know for sure, but I suspect Will's edit work because the additional neutral particle filter squeezed it in the GPU memory. @wforeman Were you using 80GB ones?

diaza:
Regarding SIGNAL_OVERLAPS_CUT, I don't think there's a problem. The filtering has been moved to before [this](https://github.com/DUNE/larnd-sim/blob/be0b6169263e53ee17363c69399625921a119562/cli/simulate_pixels.py#L748) line, which imports `detector`. Putting the filter in sim makes more sense to me anyways. 
@YifanC, this event you found is blowing up due to the number of segments, not a late segment, right? I guess that's a possibility due to the expanded pixel radius. If we can run the file with 80GB, then I think it's fine to move forward for MR6.5 and leave optimizations for future us.

YifanC:
@wforeman can confirm if running on 80GB solves the memory issue. It seems to me that the problem is not due to the late segments but the number of segments in the range of these pixels. Adjust the pixel_batch should ease the memory consumption. Perhaps setting it to 2000 is too ambitious.

wforeman:
@YifanC, I have noticed this problem too and was doing some similar tests by changing pixel batch size, even going down to 100, with no success. I ultimately had to run on 80GB nodes to get most jobs to succeed -- and even then, some still go over the limit.

The file earlier was incorrect (it was actually the one I was using to debug the problem you discovered with the 10643 tracks). Here is one that was failing from the late t0: 

`/pscratch/sd/d/dunepr/output/MiniRun6.5/run-convert2h5/MiniRun6.5_1E19_RHC.convert2h5/EDEPSIM_H5/0000000/MiniRun6.5_1E19_RHC.convert2h5.0000023.EDEPSIM.hdf5`

The problem I was finding was the printout message was telling me it filtered out the delayed segments, but I was still finding them cropping up later (see [this Slack message](https://dunescience.slack.com/archives/C073Q1023EY/p1756505576966229)).

PR #291, opened 2025-09-02
Modernize package metadata with pyproject.toml

cuddandr:
Fix a deprecation warning regarding the legacy install method by updating the package metadata and build specification (e.g. dependencies) to use `pyproject.toml` with a `setup.py` for building a dynamic dependency list.

Basically `pyproject.toml` contains most of the info, and `setup.py` is used to dynamically inspect the user's environment to install the correct version of CuPy based on the active CUDA version.

Also updated some of the metadata, such as the version, authors, and maintainers.

Let me know or edit the `pyproject.toml` to further update the package metadata. Since this is messing with the installation method, it should be tested by at least one other person.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,8 +41,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install pytest
-          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-
           pip install --editable .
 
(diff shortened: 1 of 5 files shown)


mjkramer:
Works in my tests. I added/tested a commit to remove pynvjitlink since it's no longer needed.

mjkramer (approved):

PR #293, opened 2025-10-02
Include event ID in mc_packet_assn for all packets

mjkramer:
Previously the MC event ID was being set to -1 for non-data packets (timestamps, syncs, triggers). This change makes it easier to reorder the packets event-by-event in a post-processing step (which would then make life easier for flow's event builder).

Tested:
image


diaza:
Looks good to me.
If you do something like `np.unique(event_ids)`, are all `-1`s eliminated?

mjkramer:
Good call. Indeed we have no more -1s:
image

PR #295, opened 2025-10-17
LArPix periodic reset

jaafar-chakrani:
## Motivation

This PR implements the periodic reset feature in LArPix. In detector operations, LArPix typically operates with a periodic reset to drain away slowly-accumulating charge on the front end due to e.g. leakage currents. In the past, this reset has been set to a period of ~400 microseconds, which means that it wasn't particularly important to simulate (as a drift window is ~200 microseconds). However, to mitigate far field effects, a significantly faster periodic reset has been used in more recent LArPix operations, with O(10 microsecond) periods, making it more important to take into account.

## Implementation
- The periodic reset cycles can be set from the detector properties YAML configuration file.
- The phase of the periodic reset on a given pixel is chosen randomly.
- Everytime the time tick `ic` reaches the periodic reset in [fee.get_adc_values](https://github.com/DUNE/larnd-sim/blob/10c00d4e8706a76ed5efe115a67ee262fd69fdf3/larndsim/fee.py#L538), `true_q` and `q_sum` are set to 0 and to the reset noise respectively, and the backtracking `current_fractions` is updated accordingly.
- _Temporary:_ added `qs` and `true_qs` arrays to record the changes to the charge at each time tick and save arrays of interest in an `.npz` file for debugging purposes. To be removed once this is validated.

##  Validation

So far, I ran a couple events with 25.6 and 10000.0 microseconds to compare the changes to the waveforms.

- Example 1: The periodic reset noise is reflected in the regular steps every 256 time tick. It also leads to the removal of a hit as it suppresses the long rising edge of the signal.
image 
image

- Example 2: This time, the periodic reset induces an additional early hit due to the positive charge injection.
 
image 
image

Further validation is still needed:
- [x] Look a the impact on the collected charge for shower events.
- [x] Compare the hit level distributions with and without the updates.
- [x] Compare waveforms with no periodic reset at all.
- [x] Confirm that the backtracking information is still okay for downstream analyses.

##  Remaining code updates
- [x] Add an option to disable the periodic reset (e.g. when `periodic_reset_cycles < 0`)
- [x] Remove the extra arrays used for debugging


jaafar-chakrani:
Added option to completely ignore the periodic reset if it's set < 0 and used that as the default. Below is a comparison with and without the periodic reset:
image image

Also, a comparison of the hit-level distributions (in linear and log y-axis scale): 
image

PR #296, opened 2025-10-17
FSD cosmic sim gain updates

jaafar-chakrani:
Updates to the LArPix-v2d gain following [measurements](https://indico.fnal.gov/event/71610/contributions/325972/attachments/192600/266961/v2d-gain-measurement_pixel_readout_october_16_2025.pdf) by @seg188:
- Changed the thresholds accordingly
- Changed the LArPix gain in the detector properties
image


seg188:
It looks like we have a default threshold value set around 8ke-, should we update this value to be more in line with the rest of the distribution?

jaafar-chakrani:
Using a default value more in line with the rest of the distribution now!

image

cuddandr (approved):
It runs fine for me and the plots look sensible as far as I can tell.

PR #298, opened 2025-10-28
Have MC tile layout match data

diaza:
This PR reconfigures the charge layout so that the simulation matches the data. I'll refer to the comment in the linked [PR](https://github.com/DUNE/ndlar_flow/pull/212) as validation for this PR. 
This PR should be merged in conjunction with the linked PR.

### DIFF
--- a/larndsim/config/config.yaml
+++ b/larndsim/config/config.yaml
@@ -89,7 +89,7 @@ ndlar_light:
 
 fsd:
     SIM_PROPERTIES:  cosmic_sim.yaml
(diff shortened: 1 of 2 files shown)


cuddandr:
I was able to run both larnd-sim and flow with the geometry fixes. Hand scanning a dozen or so events in the event display everything looks okay.
fsd_event_display_with_geom_fix

cuddandr (approved):
Simple enough reordering of arrays. It does in fact match the corresponding values in the nd-flow yaml, so looks good to me.

PR #300, opened 2025-11-09
Have LUT be read bottom-up, and modify the 2x2 LUTs to match true geometry

diaza:
As title says.
EDIT: [Here](https://docs.dunescience.org/cgi-bin/sso/RetrieveFile?docid=35172&filename=20251104-LUTissues.pdf&version=1)'s the talk where I went over the issues seen. If anyone missed it and would like a more detailed explanation, let me know and we can chat. 
EDIT: I'm adding everyone think might have any familiarity with the LUTs. No need to actually review if this isn't within your purview. 
To reflect the y-axis in the LUT, I used the following code:

```python
import numpy as np

def swap_y(filename, outfile):
    data = np.load(filename)['arr']
    arr = data.copy()
    arr = arr [:, ::-1, :, :]
    n_op_chan = arr.shape[-1]
    arr[..., :n_op_chan//2] = arr[..., :n_op_chan//2][...,::-1]
    arr[..., n_op_chan//2:] = arr[..., n_op_chan//2:][...,::-1]
    np.savez_compressed(outfile, arr=arr)
    return

filename = "/global/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/light_LUT/lightLUT_Mod0_06052024_time_norm.npz"
outfile = filename.replace(".npz", "_swapped_y.npz")
swap_y(filename, outfile)

filename = "/global/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/light_LUT/lightLUT_Mod123_06052024_time_norm.npz"
outfile = filename.replace(".npz", "_swapped_y.npz")
swap_y(filename, outfile)

```

Flipping for FSD shouldn't be necessary. 
The optical channel mapping will have to be remapped in flow.

### DIFF
--- a/larndsim/config/config.yaml
+++ b/larndsim/config/config.yaml
@@ -7,7 +7,7 @@ module0:
     # Path to light LUT at NERSC
     # Web portal: https://portal.nersc.gov/project/dune/data/2x2/simulation/larndsim_data/light_LUT/
     # Alternatively low granularity Mod0 based light LUT in larndsim, lightLUT.npz
-    LIGHT_LUT:       /dvs_ro/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/light_LUT/lightLUT_Mod0_06052024_time_norm.npz
+    LIGHT_LUT:       /dvs_ro/cfs/cdirs/dune/www/data/2x2/simulation/larndsim_data/light_LUT/lightLUT_Mod0_06052024_time_norm_swapped_y.npz
     LIGHT_DET_NOISE: light_noise-module0.npy
     LIGHT_SIMULATED: True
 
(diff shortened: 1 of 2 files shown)


diaza:
I've opened PRs for op channel remaps for [2x2](https://github.com/DUNE/ndlar_flow/pull/221) and [FSD](https://github.com/DUNE/ndlar_flow/pull/222).

liviocali:
@diaza I did not review the new channel maps in detail. But everything else seems sensible to me. Do you have tools to verify the new channel map? E.g. Angela at some point produced edepsim files that should deposit some charge exactly in front of the SiPMs to verify the mapping.

diaza:
> @diaza I did not review the new channel maps in detail. But everything else seems sensible to me. Do you have tools to verify the new channel map? E.g. Angela at some point produced edepsim files that should deposit some charge exactly in front of the SiPMs to verify the mapping.

If @Angela-White has that ready to send, that would be great! Otherwise I can try to create my own, I've done something similar for 2x2

diaza:
@liviocali 
I've attached two gifs, showing how the light deposition changes as a function of charge deposition position in the FSD. The first shows for the existing code, the other for my modifications. You can see that the depositions in the second gif look more sensible. 
EDIT: Replacing with better gifs.
develop_fsd
feature_fsd

diaza:
For good measure, I'm showing below two gifs with a similar comparison as above, but in the 2x2. This comparisons should show that the modifications made to larndsim, the LUT, and the op channel mapping in 2x2 result in the same output. The top gif is for the existing code, and the second gif is for the updated. 
develop_2x2
feature_2x2

cuddandr (approved):
The GIFs are enough validation for me. Everything looks good to me.

PR #301, opened 2025-11-12
Use dynamic pixel batches based on num of segments

mjkramer:
Previously, each sub-batch (of segments) corresponded to a fixed-size region of pixels. As a result, the number of segments per sub-batch could vary wildly. With this, the pixel region is dynamically sized in order to get the number of segments as close as possible to the (new) `segment_batch_size` parameter. The `pixel_batch_size` parameter is gone.


diaza:
Looks good to me!

diaza (approved):

PR #302, opened 2025-11-21
NDLAr Config Choices for MiniProdN5 Nominal (Sample A)

alexbooth92:
Changing the `max_radius` to 4 and switching on light backtracking for NDLAr.


cuddandr (approved):
Looks good to me. How concerned are you about running out of GPU memory with the light backtracking and increased max radius? I'm nearly done with halving the light LUT memory size (famous last words) which would cut ~9.5 GB from the simulation.

alexbooth92:
Thanks Andrew! Matt was able to run a successful test with a single file. I guess we'll get a handle on the failure rate due to memory as we scale up 😅

An extra 10 GB of headroom would be very nice!

mjkramer (approved):

PR #311, opened 2026-01-07
Pixel response file with shield plane at 900V for ND-LAr

jaafar-chakrani:
# Summary

I'm adding here the pixel response file with shield plane. This file is obtained from Jiangmei's [simulation](https://github.com/jiangmei-yang/LArPix_Field_Simulation/tree/main/LArPix_v3_5x5_shield) at 900 V for the 2x2 drift length (~30 cm), and padded with zeros to match the ND-LAr drift length (~47 cm). This should be fine because the induced current between 30 and 50 cm is multiple magnitudes smaller in the shield-plane case (see bottom plot).

# Validation

## Values and shapes
```python
import numpy as np

new_response = np.load('response_37_v2d_shield900V_fsd_ndlar_full.npz')

print("Response file w/ shield plane:")
print(" * Response shape: ", new_response['response'].shape)
print(" * Time tick: ", new_response['time_tick'])
print(" * Drift length: ", new_response['drift_length'])
print(" * Bin size: ", new_response['bin_size'])

old_response = np.load('response_37_v2d_fsd_ndlar_full.npz')

print("Response file w/o shield plane:")
print(" * Response shape: ", old_response['response'].shape)
print(" * Time tick: ", old_response['time_tick'])
print(" * Drift length: ", old_response['drift_length'])
print(" * Bin size: ", old_response['bin_size'])
```
Output:
```
Response file w/ shield plane:
 * Response shape:  (45, 45, 2975)
 * Time tick:  0.1
 * Drift length:  46.788
 * Bin size:  0.0372
Response file w/o shield plane:
 * Response shape:  (45, 45, 5950)
 * Time tick:  0.05
 * Drift length:  46.788
 * Bin size:  0.0372
```
Notice that the time tick in the with-shield file is 100 ns vs. 50 ns in the no-shield one, but this is taken care of properly in larnd-sim (e.g. [here](https://github.com/DUNE/larnd-sim/blob/e3e5e7600da4170a3720ad2caa44bc3a08a8caf8/larndsim/consts/detector.py#L371) and [here](https://github.com/DUNE/larnd-sim/blob/e3e5e7600da4170a3720ad2caa44bc3a08a8caf8/larndsim/detsim.py#L45-L50)).

## Induced current at different x, y positions
image

## Region padded with zeros
image


jaafar-chakrani:
Summarizing Dan's comment here: the suppression of the induced current at the far field (as shown in the last plot above) appears to be very high, probably due to idealized assumptions in the Garfield simulation, so it needs some additional thoughts/cross checks.

PR #314, opened 2026-02-17
Skip the periodic reset during the hit cycle

jaafar-chakrani:


cuddandr:
Checked with known charge deposits with and without the fix (ADC noise and diffusion turned off). The fix recovers all the pixels that were missing a hit achieving 100% pixel efficiency and the tail toward lower dataword/charge/energy is less pronounced. With the fix the periodic reset can still cause charge loss, but now is less of an impact.
```
PRC before fix:
- Total packets: (390463,)
- Unique pixels: (390463,)
PRC after fix:
- Total packets: (409600,)
- Unique pixels: (409600,)
```
charge_injection_prc_compare

The same plot for a simple muon particle gun before and after the fix (using FSD Sim v4 nominal thresholds) and numbers for pixels/packets. Fewer packets/pixels are lost and fairly uniform across the distributions.
```
PRC before fix:
- Total packets: (269264,)
- Unique pixels: (160488,)
PRC after fix:
- Total packets: (291473,)
- Unique pixels: (167719,)
```
fsd_gun_nomThr_compare

PR #315, opened 2026-02-24
Far-field signal calculation

jaafar-chakrani:
# Overview

This PR introduces far-field dipole signal simulation into the charge readout pipeline, enabling complete electric field response calculation for detector characterization. Far-field signals from voxelized ionization charge are now computed and added to the pixel response alongside near-field contributions.

# Details

## Architecture

The far-field simulation is integrated into the pixel batch processing loop as follows:

```
For each pixel batch:
  1. Calculate near-field signals (existing)
  2. Accumulate pixel signals from all contributing tracks
  3. [NEW] Add far-field dipole contribution per TPC
  4. [NEW] Handle induction-only pixels (far-field only)
  5. Simulate electronics response (ADC digitization)
```

#### 1. **Voxelization (Per-Event-Batch)**
- Tracks are voxelized once per event batch to avoid cross-TPC contamination
- Voxel positions: $(x_v, y_v, z_v)$ in detector coordinates
- Voxel charges: summed ionization per voxel
- Separate caches maintained per TPC

```python
for tpc_idx in active_tpc_indices_all:
    voxel_indices, voxel_charges, ... = gpu_voxelize(tpc_tracks)
    voxel_cache[tpc_idx] = {x, y, z, q}  # GPU cupy arrays
```

#### 2. **Pixel Classification**
- Pixels categorized as **collection**, **neighbor**, or **induction-only** (right now, collection and neighbor are treated in the same way)
- Classification cached per TPC to avoid recomputation
- Used to identify pixels requiring only far-field treatment

```python
classification_cache[tpc_idx] = pixel_classifier.classify_pixels(
    tracks, plane_id=tpc_idx
)
```

#### 3. **Far-Field Signal Calculation**

**Mixed Pixels** (Near + Far-field):
- Computed per TPC for pixels with track contributions
- Dipole signals added after near-field accumulation
- Handled in per-pixel-batch loop

**Induction-Only Pixels** (Far-field only):
- Processed in separate loop after pixel batches
- Uses classifier's pre-computed pixel coordinates
- Avoids redundant classification per batch

#### 4. **Time Sampling & Digitization**
- Far-field signals computed on same time grid as near-field

# Validation

See presentation [here](https://docs.dunescience.org/cgi-bin/private/ShowDocument?docid=35905).


jaafar-chakrani:
Additional validation and known issues [here](https://docs.dunescience.org/cgi-bin/private/ShowDocument?docid=36223).

mjkramer:
OK, after some reorganization and fixes (#322, #324), I think this is ready to be merged, even though physics validation is still ongoing. Given the large size of this PR, and other refactoring work going on in parallel, merging now makes sense to me. Further changes to the FFE sim will be the subject of future PRs.

By default, the FFE sim is disabled, giving idential output to the `develop` branch (confirmed with `cli/diff_files.py` for the FSDCube case). It can be enabled via the `farfield_enabled` simulation property, and the mode can be selected by setting `farfield_mode` to `segments` (the default) or `voxels`.

The bulk of the new code is confined to `larndsim/far_field` and `larndsim/consts/ff_induction`, where it does nothing if FFE is disabled. However, some changes have been made to `consts.detector` and `consts.sim`, and more importantly, to `simulate_pixels.py`.

In the process of preparing `simulate_pixels.py`, I did a bit of "extra" refactoring, namely, defining two new top-level functions containing the code from `save_results` and the code surrounding the call to `get_adc_values` (which we now call a second time for FF-only pixels). I was hoping `run_simulation` would actually get shorter when all was said and done, but alas, it's still about 40 lines longer than before.

cuddandr:
Confirmed that the FFE branch runs fine for me using FSD cosmic simulation, and that the case of FFE off returns identical results to the current `develop` branch. Shown below are two plots of the charge packet datawords for `develop` and this branch for FFE on and off. The difference between FFE on and off is minimal, but this is expected and encouraging since the FSD cosmic events are mostly muons and occasional delta rays.

ffe_on_off_comparison
fee_off_develop_comparison

mjkramer:
Pushed a fix for the typo you found in the image charge locations (was `2*n*l`, now `(2**n)*l`.

Compared to before, in "segments mode", there are no changes to the `packets`, but there are numerous small changes (< 1e-4 typically) to the `fraction` and `fraction_traj` of the `mc_packets_assn`, which is unsuprising given that the total charge (i.e. the denominator) has changed slightly due to the modified FFE calculation.

In "voxels mode", the `fractions` are affected similarly, and there are also three packets (out of 25k) where the dataword shifted by ~5 counts. It's also unsurprising that only "voxels mode" actually saw a change in the `packets`, since this mode (currently) produces stronger FF effects, for reasons still under investigation.

cuddandr (approved):
Everything looks good as far as I can tell from reading the code. I have not spotted any other mistakes, but I'm also not very familiar with all the FFE code.

Anyway, I agree that we should go ahead and merge it since it the segments mode looks like it's doing something reasonable and further developments will be easier as part of the main branch.

PR #317, opened 2026-03-13
Option to use segments directly instead of voxels for far-field calculation

jaafar-chakrani:
This PR adds a new far-field simulation mode to evaluate the impact of voxelization on induced-current modeling. In addition to the existing voxel-based far-field path, the code can now compute far-field induced current directly from track segments outside the near-field region (after breaking them down into 5-mm subsegments). This mode is enabled via: `--farfield_enabled --farfield_mode segments`

The PR also reduces `MAX_RADIUS` to 2 pixels, so the dipole-based far-field calculation is applied immediately beyond that distance. The motivation is to recover contributions that can be missed in the detailed LUT-based path due to its cap on the number of segments contributing per pixel. The far-field dipole calculation does not have this per-pixel segment cap, which improves the response for denser events and at larger distances.

### DIFF
--- a/cli/simulate_pixels.py
+++ b/cli/simulate_pixels.py
@@ -174,7 +174,8 @@ def run_simulation(input_filename,
                    rand_seed=None,
                    compression=None,
                    save_memory=None,
-                   farfield_enabled=False):
+                   farfield_enabled=False,
+                   farfield_mode='segments'):
     """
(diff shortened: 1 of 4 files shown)


mjkramer:
Merging feature branches to prepare for merging into `develop`

PR #318, opened 2026-04-01
Fix the padding of the pixel response

jaafar-chakrani:
The field response is represented as a 3D array with dimensions *(x, y, t)*. When the configured drift distance exceeds that covered by the field response, the response must be extended along the time axis to represent the additional drift.

The current implementation:

```python
cp.pad(response, abs(drift_ticks_diff))
```

pads all axes symmetrically (both before and after each dimension), which is incorrect for this use case (see [documentation](https://docs.cupy.dev/en/latest/reference/generated/cupy.pad.html)).

Instead, we should only pad the time axis, and only on the left side:

```python
cp.pad(response, ((0, 0), (0, 0), (abs(drift_ticks_diff), 0)))
```

This preserves the spatial dimensions along the pixel plane *(x, y)* and correctly applies a shift in time by adding zeros at the beginning of the waveform.

### DIFF
--- a/larndsim/consts/detector.py
+++ b/larndsim/consts/detector.py
@@ -432,10 +432,10 @@ def load_response(response_file):
 
     if drift_ticks_diff > 0: # response is too long, chop
         response = response[drift_ticks_diff:]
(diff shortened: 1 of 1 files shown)


cuddandr:
The fix works fine for me running the FSDCube configuration with its correct (longer) drift length. The output is very slightly different compared to using the shorter FSD drift, but that is unsurprising as more segments can now be processed in the larger volume.

cuddandr (approved):

